PR #5, opened 2020-10-10
Add some basic movement stuff, placeholder models

TheRyeGuyWhoWillNowDie:


KinematicBody2D.gd
```diff
@@ -0,0 +1,21 @@
...
+func get_input():
+	velocity = Vector2()
+	if Input.is_action_pressed('right'):
+		velocity.x += 1
+	if Input.is_action_pressed('left'):
+		velocity.x -= 1
+	if Input.is_action_pressed('down'):
+		velocity.y += 1
+	if Input.is_action_pressed('up'):
+		velocity.y -= 1
```
monban, line 16: I believe these are backwards.
```suggestion
	if Input.is_action_pressed('right'):
		velocity.x -= 1
	if Input.is_action_pressed('left'):
		velocity.x += 1
	if Input.is_action_pressed('down'):
		velocity.y -= 1
	if Input.is_action_pressed('up'):
		velocity.y += 1
```

KinematicBody2D.gd
```diff
@@ -0,0 +1,21 @@
...
+	velocity = velocity.normalized() * speed
```
monban, line 17: Setting the velocity explicitly like this means there is no acceleration, character immediately starts moving at max speed. This is contrary to what @cukmekerb implemented in #2. Not saying we should prefer one over the other, I just hate to see duplicated efforts.

KinematicBody2D.gd
vincens2005: It isn't

MGThePro: Can confirm, with the arrow keys I am walking in the correct direction

monban: Oh hey you guys, I just looked more closely at the scene and it turns out I'm an idiot, since the bg and character have the same graphic I was not noticing that the camera is locked to the character, and interpreted the bg moving as the character. I formally rescind this suggestion.

KinematicBody2D.gd
```diff
@@ -0,0 +1,21 @@
+extends KinematicBody2D
+
+export (int) var speed = 200
```
MGThePro, line 3: It's probably a good idea to start using global constants for values that users might change on a per lobby basis.
Not sure how it's commonly done in GDScript, but a quick google search suggested creating some .gd file (like lobbySettings.gd, for example), setting the variables there, and then using lobbySettings.playerSpeed from the file we created earlier.

TheRyeGuyWhoWillNowDie:
Wow, Im dummy. I didn't realise the game had stuff for it yet. Closed

PR #19, opened 2020-10-11
added german language support

Wandang:
This adds german language support


Wandang:
i will create a new pr from a feature_branch since that is the appropriate workflow

PR #20, opened 2020-10-11
Feature german translation

Wandang:
* Adds german translation file and translation itself


MGThePro:
Translation looks good :+1:

PR #31, opened 2020-10-11
a fix for the strange invisible host bug, properly sized "join game" dialogue

TheRyeGuyWhoWillNowDie:
fixes the console error-spam when the server tries to rpc itself, host still invisible :(


macmv:
How does this fix things? I would think it should call `player_moved` directly, not `rpc("player_moved")`, as that will call `player_moved` on all other clients

TheRyeGuyWhoWillNowDie:
> How does this fix things? I would think it should call `player_moved` directly, not `rpc("player_moved")`, as that will call `player_moved` on all other clients

This fixes some of the issues because it bypasses sending to the server if you are the server. It doesn't fix the invisible host issue, but it does stop the error-spam

PR #46, opened 2020-10-12
allow use of analog stick on a controller to move

darshadi:


MGThePro:
Is this compatible with Xinput or Dinput?

darshadi:
I used godot's input system so it should work with everything that godot works with

PR #60, opened 2020-10-13
Implement WebSocket Networking

TheSecondReal0:


src/Scripts/network.gd
```diff
@@ -7,5 +7,67 @@ enum Connection {
 }
 
 var connection = Connection.LOCAL
-var host: String = ''
+var host: String = '' #not sure why this is needed or a string, server network ID will always be 1
```
monban: This was intended to be used as the hostname of the server. It is set to localhost when connected locally (see [PlayGame.gd line 17](https://github.com/opensuspect/opensuspect/blob/d8bc5b247e45ce3d7fb701c66429039bf1416e1d/src/Scripts/PlayGame.gd#L17)).

TheSecondReal0: I see, maybe change the name of that variable to be hostName?

monban: I'm good with that.

monban: Would you please go ahead an change the variable name and remove the comment, so we can merge? Thank you.

tooManyArguments:
Also might have to rebase since we changed file / dir structure:


Conflicting files
src/Scripts/network.gd

(all is now lowercase)

daniel071 (approved):
Just tested it and it works great:
image

Consider reducing the console spam:
image

When someone disconnects, this error occurs:
image

PR #68, opened 2020-10-13
Fix file names (all lower case and rename kinematicbody2d.gd to player.gd)

darshadi:


darshadi:
Will reopen when it works

PR #69, opened 2020-10-13
Fix file names (all lower case and rename kinematicbody2d.gd to player.gd) and change to gles2 and make controller input better

darshadi:


tooManyArguments:
in my project settings > autoload it is now referencing `res://Scripts/network.gd`
would you please double check on this?

tooManyArguments:
Tested again - OK

PR #78, opened 2020-10-14
New settings system.

GrbavaCigla:
New settings system (currently supports only switches)
What works:
- Autosaving 
- Autoloading
- Autogenerating settings screen
- Events
- Switches

What doesn't work:
- Rest of the widgets
- Colorblind (made shader that will come in next pr)
- Ugly UI, but better then before

How to add options:
To add new setting add new element to list like this (currently):
```gdscript
var settings = [
	Setting.new(true, SettingType.SWITCH, tr("Fullscreen"), "toggle_fullscreen"),
	Setting.new(false, SettingType.SWITCH, tr("Colorblind mode"), "toggle_colorblind"),
]
```
First argument is default value
Second argument is node type
Third argument is label displayed
Fourth argument is function name

Next add function:
```gdscript
func toggle_fullscreen(setting):
	OS.window_fullscreen = setting.value
```

Why autosaving works?

Switching CheckButton on or off triggers wrapper function what modifies setting in array and prepares everything for given function and it also **saves** that setting to a file.

Why do we need this?
We need this because this is more modular, cleaner and easier to use

### DIFF
--- a/src/scenes/ui/settings.tscn
+++ b/src/scenes/ui/settings.tscn
@@ -3,116 +3,11 @@
 [ext_resource path="res://scripts/settings.gd" type="Script" id=1]
 
 [node name="Settings" type="ScrollContainer"]
-anchor_left = 0.5
-anchor_top = 0.5
-anchor_right = 0.5
-anchor_bottom = 0.5
-margin_left = -512.0
-margin_top = -300.0
-margin_right = 512.0
-margin_bottom = 300.0
+anchor_right = 1.0
+anchor_bottom = 1.0
 size_flags_horizontal = 3
 size_flags_vertical = 3
 script = ExtResource( 1 )
 __meta__ = {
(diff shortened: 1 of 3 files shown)


monban:
Tested and settings correctly persist through multiple runs on my system (I use Arch btw). I **love** the modular way it's laid out. Only one request before this PR is accepted: there seems to be no way to back out of the settings menu? The BACK button is missing / not displayed for me.

GrbavaCigla:
> Tested and settings correctly persist through multiple runs on my system (I use Arch btw). I **love** the modular way it's laid out. Only one request before this PR is accepted: there seems to be no way to back out of the settings menu? The BACK button is missing / not displayed for me.

Forgot that one... How do I commit to pr? I just commit and click pull again or something else?

GrbavaCigla:
OK, fixed it now

PR #81, opened 2020-10-15
Use the players' intended movements for animations instead of the velocities

vainiovano:
The movement value fits also well for a client-server multiplayer architecture. The player can send the movement value to the server and the server can validate it easily.

The players can now walk against walls.

walking against a wall

Please only push functioning code! The first commit of this PR took way too much time to make. Bisecting is almost impossible because there are so many untested commits that don't work.

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -41,7 +41,6 @@ func _player_connected(id):
 		rpc_id(id, "player_join", new_player.id)
 
 	players[id] = new_player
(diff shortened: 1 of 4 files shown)


tooManyArguments:
There was a merge of new websocket code, could you please check on this conflict?

TheSecondReal0:
I have since tested and #83 did break this

vainiovano:
Rebased. I edited the commit message of the first commit to reflect the current state.

PR #87, opened 2020-10-16
apply law of demeter to network node

monban:
Decouple the menu from the network code, allowing for more abstract and less brittle interactions.


tooManyArguments:
please rebase and check on conflicts.

PR #89, opened 2020-10-16
Colorblind mode

GrbavaCigla:
Added colorblind shader, added colorblind to settings, made dummy function in settings. When settings system gets comboboxes, it will support all kinds of color blindnesses


TheSecondReal0:
I would recommend putting the shaders folder in assets/common instead of in assets, just for the sake of organization

GrbavaCigla:
Fixed

tooManyArguments:
Probably due to new merges, please rebase and check again (conflict)

GrbavaCigla:
Will do

GrbavaCigla:
Resolved conflicts

vainiovano:
How would swapping the colors help people with color blindness? Wouldn't it just move the problematic colors?

GrbavaCigla:
I copied the shader from here: https://www.reddit.com/r/gamedev/comments/9u1y6f/adding_colorblind_accessibility_with_an/
It helps, but it isn't ideal

vainiovano (approved):
Really needs further testing with different colors but approving for now

PR #94, opened 2020-10-17
Added the new naming system

TheRyeGuyWhoWillNowDie:
The chat also now has the names of players, instead of just their ids


vainiovano:
Could the players list be shared in some sort of way so that `chatbox.gd` could read it? The players wouldn't have to send their names all the time.

vincens2005 (dismissed):

GrbavaCigla (approved):

PR #95, opened 2020-10-17
Added a maps system and fixed scaling in main

### LINKED ISSUE #85: Fix weird scaling in main.tscn
Right now, the root node of main.tscn is scaled by a factor of .05. I think this is because the textures are so big that they would take up the entire screen if they weren't massively scaled down, but this is a poor solution. We need to decide on standards for image size and/or move a similar solution elsewhere, maybe by scaling down the actual Sprite node instead the entire scene.

TheSecondReal0:
Organized to make it easier to expand, added a system to switch and keep track of maps, fixed weird scaling in main.tscn  (#85), expanded the name label so that long names are still centered.

### DIFF
--- a/src/assets/autoload/gamemanager.gd
+++ b/src/assets/autoload/gamemanager.gd
@@ -1,7 +1,10 @@
 extends Node
 
+#this script will sync game events, like round resetting and meetings
 var ingame: bool = false
 
+#signals that help sync the gamestate
+#can be connected to from anywhere with GameManager.connect("<signal name>", self, "<function name>")
 signal resetLobby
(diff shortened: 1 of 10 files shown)


monban (commented):
I'm getting a crash when the second player joins, may I get a second review on this to verify it's not just me? @TheSecondReal0 is it crashing for you?

GrbavaCigla:
Crashed for me too
Also game that doesn't crash, scales the sprite across the whole screen

monban (approved):
Re-tested, crash is gone :+1:

PR #97, opened 2020-10-18
added client/server version check

TheRyeGuyWhoWillNowDie:
not enforced right now, but warns you in console. if you run into any crashes while testing pull requests PLEASE check the console, it will tell you if the version doesnt line up.


TheSecondReal0 (approved):
Tested and works

PR #108, opened 2020-10-18
Add some basic automatic role assignment, using some RNG

TheRyeGuyWhoWillNowDie:
Generate a random number 1-10, if the number is 8 or greater, set the player to be an intruder. current limit is 2 intruders max. Currently no way to force an intruder minimum, but that can be added soon.

### DIFF
--- a/src/assets/autoload/gamemanager.gd
+++ b/src/assets/autoload/gamemanager.gd
@@ -2,7 +2,6 @@ extends Node
 
 #this script will sync game events, like round resetting and meetings
 var ingame: bool = false
-var isgamestarted: bool = false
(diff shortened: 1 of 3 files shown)


vainiovano:
I'd like to be able to select a fixed number of impostors. The code could work by getting the first elements of a shuffled array.

TheRyeGuyWhoWillNowDie:
> I'd like to be able to select a fixed number of impostors. The code could work by getting the first elements of a shuffled array.

Max intruders can be set as a variable that can be edited in the menu, not hard to implement

GrbavaCigla (approved):

PR #112, opened 2020-10-18
Add lobby and improve map system

### LINKED ISSUE #106: Map management system
A system to manage and store info about maps. This info would be the name, file directory, what tasks are available on it, etc.. Features to include:
- Keep track of every map in the game and info related to said map
- Switch maps

### LINKED ISSUE #109: Support for non 16:9 Aspect Ration
Currently the game is forced to 16:9 by black bars, either on top and bottom or left and right.
It would be good to support more aspect ratios.

### LINKED ISSUE #110: UI doesn't adapt to different screen sizes
Changed the screen resolution to 1920x1080 and since many of the ui nodes aren't properly set to centered/full rect/under a center container etc many things are moved to either off screen or to the side whenever screen resolution isn't 1024x600. This will become a big issue in the future if it isn't taken care of soon.

TheSecondReal0:
Added a lobby map and made it so the map switches when the game starts. UI scaling is also improved.

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -134,8 +134,10 @@ func _on_startgamebutton_gamestartpressed():
 			intruders = intruders + 1
 		rpc_id(other_id,"startgame",isintruder)
 		isintruder = false
+	$maps.switchMap("test")
 	emit_signal("clientstartgame")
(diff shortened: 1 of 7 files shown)


monban (approved):
Code looks solid, tested the game, seems to be working fine.

PR #114, opened 2020-10-19
Added Player Visibility range

vincens2005:
players can only see each other from certain distance. To change, simple change the ViisibleArea node's scale in player.tscn


monban (approved):
Code looks good, tested game and it seems to be working fine.

PR #116, opened 2020-10-19
Improve map info storage

TheSecondReal0:
Nothing is functionally different, it shouldn't even cause version mismatch crashes. All this does is format the map dict correctly so it can be expanded to store more info about each map.


TheRyeGuyWhoWillNowDie (approved):
nice

PR #118, opened 2020-10-19
Modify player controller to handle simultanous opposite input

ghost:
- movement direction now equal to net keypress strength


monban (approved):
I believe this will also making playing on analogue controller better as well. :+1:

PR #121, opened 2020-10-19
Changed README just a little

jgx0:


log.txt
```diff
@@ -0,0 +1,2 @@
+2020/10/19 12:26:16 Micro started
+2020/10/19 12:26:16 monokai-dark is not a plugin
```
vainiovano: What is this?

TheRyeGuyWhoWillNowDie (dismissed):
kind of weird but ok

log.txt
jgx0: sorry its my git log you can delete that

monban (approved):
Weird flex but okay.

PR #122, opened 2020-10-19
Moved names and cleaned up main.gd

### LINKED ISSUE #98: Consolidate networking stuff into network.gd
Right now connection handling is split between main.gd and network.gd, when ideally it should all be in network.gd. This includes names, version checking, etc..

TheSecondReal0:
Moved the name system into network.gd and generally cleaned up main.gd. I removed a lot of unnecessary networking stuff from main, and created a new signal that network.gd emits when a player connection is done being handled. Consolidation of network stuff isn't complete with this, so don't close #98.


monban (approved):
I noticed that when the server transitions to the game map (test) the client stays in the lobby (though they can still see each other). I think this existed prior to this PR, so I'm going to approve it and put in a bug that we can fix.

PR #126, opened 2020-10-20
Fix FSM syncing and improve role assignment

### LINKED ISSUE #100: Role assignment system
A basic system to assign roles to players. Features to include:
- Easily expandable with more roles
- Scales by the amount of players in the lobby
- Customizable (how many of each role should be spawned/what percentage of players should be x role, etc.)

### LINKED ISSUE #123: Sync FSM across network
Right now the FSM only works server side, meaning game states aren't synced.

### LINKED ISSUE #124: Client does not change map
When the server presses the "Start Game" button, the server changes maps, but the client does not. The client remains on the "Lobby" map. Both players can still see each other.

TheSecondReal0:
The FSM in gamemanager.gd is now synced across the network. The role assigning system now allows an infinite number of roles and 2 options for how to assign roles (percent players x role or y players x role). Role assignment was moved out of main.gd and into playermanager.gd. I wasn't sure where to put role assignment, so I made sure it would be easy to cut and paste somewhere else.

### DIFF
--- a/src/assets/autoload/gamemanager.gd
+++ b/src/assets/autoload/gamemanager.gd
@@ -14,9 +14,9 @@ const TRANSITIONS = {
 signal state_changed
 
 func _ready():
+	set_network_master(1)
 	get_tree().connect("connected_to_server", self, "_on_connected")
 	Network.connect("server_started", self, "_on_connected")
-
 
 func transition(new_state) -> bool:
 	print("attempting to transition gamestate from ", state, " to ", new_state)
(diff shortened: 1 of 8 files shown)


monban (approved):
Reviewed code, solid design, good implementation. Ran binaries, tests conducted:

1. Ran binary twice
1. Created server on one, joined from the other
1. Tested chat both directions
1. Started game from server, observed both clients to change state
1. visibility still works as expected, no other weirdness

-

PR #127, opened 2020-10-20
Use puppet instead of remote in other_player_moved

vainiovano:
The puppet keyword ensures that nobody is impersonating the server.


TheSecondReal0 (approved):
Code looks good, tested with two exported clients and it works perfectly.

PR #128, opened 2020-10-20
Fix role assignment

TheSecondReal0:
Using the break keyword was breaking out of the entire for loop, when it should only be breaking out of one iteration of the for loop. "Break" is replaced with "continue". This allows role assignment to keep going if an enabled role is not a real role.


monban (approved):
Sorry for the delay, been a busy day for me!

PR #131, opened 2020-10-21
Add UI system

TheSecondReal0:
Created a UI system that might go a bit overboard with decoupling, by even decoupling buttons from the menus they're supposed to open. Also adds a pause menu, which can be used to leave or quit the game and opens with the ui_cancel keybind (probably esc). 

How the UI system works:
1. Script sends a signal to uimanager.gd to open a menu with UIManager.open_menu(menuName)
2. uimanager.gd emits a signal to open that menu, which uicontroller.gd (attached to a CanvasLayer in main.tscn) picks up
3. uicontroller.gd decides whether or not it needs to reload the menu (delete from scene tree and reinstance) or not
4. uicontroller.gd now knows that that UI element is instanced and is in the scene tree, and so calls open() in it
5. UI element does whatever it does to open, maybe checks if it's possible that it can open, and then tells uimanager.gd that it opened with UIManager.menu_opened(menuName)
6. When UI element closes, it calls close()
7. close() will do whatever the menu needs to do to close and then tells uimanager.gd that it closed with UIManager.menu_closed(menuName)


monban (approved):
new ui code looks beautiful, tested game and everything working fine.

PR #133, opened 2020-10-21
Switched many nodes to YSort

TheSecondReal0:
Switched many nodes in main.tscn to be YSort, and switched the map scene roots to be YSorts. This allows Y sorting to work as intended.


TheRyeGuyWhoWillNowDie (approved):
Looks good

PR #137, opened 2020-10-23
Fix various bugs

TheSecondReal0:
When the connection is severed peers and names are cleared, the game can now switch back and forth between lobby and normal freely, start game button now toggles game state between lobby and normal.

### DIFF
--- a/src/assets/autoload/scenemanager.gd
+++ b/src/assets/autoload/scenemanager.gd
@@ -6,7 +6,9 @@ func _ready() -> void:
 	GameManager.connect('state_changed', self, '_on_gamestate_change')
 	
 func _on_gamestate_change(old_state, new_state) -> void:
-	if (new_state == GameManager.State.Lobby):
+	if new_state == GameManager.State.Lobby:
+		if old_state == GameManager.State.Normal:
+			return
 		print("gamestate just changed to Lobby, loading scene...")
 		get_tree().change_scene("res://assets/main/main.tscn")
 	if new_state == GameManager.State.Start:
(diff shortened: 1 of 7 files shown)


src/assets/autoload/scenemanager.gd
```diff
@@ -6,7 +6,9 @@ func _ready() -> void:
 	GameManager.connect('state_changed', self, '_on_gamestate_change')
 	
 func _on_gamestate_change(old_state, new_state) -> void:
-	if (new_state == GameManager.State.Lobby):
+	if new_state == GameManager.State.Lobby:
+		if old_state == GameManager.State.Normal:
```
monban, line 10: I knew there was a reason I decided to pass old_state in the signal. Congratulations on being the first person to actually use it (as far as I know)/

src/assets/player/player.gd
```diff
@@ -40,10 +40,18 @@ func roles_assigned(playerRoles: Dictionary):
...
-		$Label.set("custom_colors/font_color", Color(1,0,0))
-	if role == "detective" and PlayerManager.ourrole == "detective":
-		$Label.set("custom_colors/font_color", Color(0,0,1))
+	match role:
+		"traitor":
+			if PlayerManager.ourrole == "traitor":
+				setNameColor(Color(1,0,0))
+		"detective":
+			#not checking if our role is detective because everyone should see detectives
+			setNameColor(Color(0,0,1))
```
monban, line 49: I don't think this is working right now, when I join two players, the detective sees himself as blue, but the other player sees him as white still. Don't think this is worth holding up the PR for, but do you want to look at it?

monban (approved):
Approving with notes, your call if you want to fix it now or merge now and you or someone else can circle back on this (largely unrelated) problem down the line.

PR #138, opened 2020-10-23
Add badges to readme

jgx0:


monban (changes requested):
Honestly, I'd like to see this moved to the top to replace the "jelly bean man" we have there as temporary art. At least until such time as the Branding Team can supply us with something like an official logo.

monban (approved):

PR #139, opened 2020-10-23
Improved light system for opensuspect

zapzoop0099:
This is an updated light system for the opensuspect it will now make the environment a dark little bit but the player will be able to see those elements  in the game(like walls,buttons,doors) and can see map element more clearly but cannot see other people as i the among us and also the dark.png is not needed anymore so someone else is free to remove that.Please review it before merging.


monban (approved):
Looks good.

PR #140, opened 2020-10-23
Added the foundations of automatic task assignment.

TheRyeGuyWhoWillNowDie:
Currently just RNGs all entries in an array, but it is a good start.

### DIFF
--- a/src/assets/autoload/playermanager.gd
+++ b/src/assets/autoload/playermanager.gd
@@ -6,7 +6,8 @@ extends Node
 var inMenu = false
 var ourrole
 var ournumber
-
+var tasks = [-1]
+var localtasks
 #vars for role assignment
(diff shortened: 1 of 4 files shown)


src/assets/autoload/playermanager.gd
```diff
@@ -16,13 +17,19 @@ var roles: Dictionary = {"traitor": {"percent": float(2)/7, "amount": 1, "critic
...
 
 func _ready():
 	set_network_master(1)
 	GameManager.connect("state_changed", self, "state_changed")
-
+func assigntasks():
+	for id in Network.peers:
+		for task in tasks:
+			if task == -1:
+				rng.randomize()
```
vainiovano, line 30: There should be no need to call randomize multiple times. It could make the numbers less random.

monban (approved):

PR #141, opened 2020-10-23
Use YSort on the alien

vainiovano:
Editing the z index works on top of YSorting.


TheSecondReal0 (approved):
Looks good, alien is no longer basically a rug

PR #142, opened 2020-10-24
QoL improvements to some elements

nicemicro:
1. The game can be started in both join and create case by pressing enter in the name box.
2. When the chatbox is opened, the chat input box gets focused on automatically.


TheSecondReal0 (approved):
Code looks good, works perfectly

PR #144, opened 2020-10-25
Added the new task handling logic

TheRyeGuyWhoWillNowDie:


src/assets/autoload/playermanager.gd
```diff
@@ -25,11 +26,23 @@ func _ready():
 	GameManager.connect("state_changed", self, "state_changed")
 func assigntasks():
 	for id in Network.peers:
-		for task in tasks:
+		localtasks = tasks
+		print("poo",localtasks)
```
monban, line 30: I'm all about debugging output, especially at this point in debugging, but do you think we could be a little more clear about the purpose of the message?

src/assets/tasks/dev/Button.gd
```diff
@@ -0,0 +1,24 @@
...
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+#func _process(delta):
+#	pass
+
+
+func _on_Button_pressed():
+	print(PlayerManager.assignedtasks[0])
+	if PlayerManager.assignedtasks[0] == 0:
+		print("yeth")
```
monban, line 22: "yeth"?

monban (changes requested):
Please make the debugging messages a little clearer as to what they are logging.

PR #145, opened 2020-10-25
Improved standbutton

TheSecondReal0:
The standbutton now has a z index of -1, so you are always standing on top of it. It also now behaves correctly when multiple players walk over it at the same time.


monban (approved):
:+1:

PR #146, opened 2020-10-25
Various UI improvements

TheSecondReal0:
Chat button new reappears when lobby is entered, and the "OpenSuspect" title text no longer jumps around when navigating the main menu.


monban (approved):
:+1:

PR #153, opened 2020-10-30
(finally) added the task assignment

TheRyeGuyWhoWillNowDie:
Also made the clock task require being assigned in order to open


TheSecondReal0 (approved):
Looks good, works flawlessly

PR #154, opened 2020-10-31
Lobby map and misc lighting stuff

TheSecondReal0:
* Created horizontal and vertical wall scenes to be copy and pasted around to create basic maps fast
* Added the start of a train in the lobby map, still need a moving background for that scene
* Fixed the weird lighting stuff in player.tscn
* Your own player sprite can no longer be hidden due to light occlusion


TheRyeGuyWhoWillNowDie (approved):
LGTM

PR #155, opened 2020-10-31
Move players' and alien's position to be inside their collision area

vainiovano:
This will ensure the correct YSort order.


monban (approved):
:+1:

PR #166, opened 2020-11-03
Player interaction

TheSecondReal0:
* Created an interactpoint prefab
* Walking near an interactpoint will bring up a UI button to interact with it
* Cannot interact with them through walls
* They can be used to interact with nodes and to open UIs


TheRyeGuyWhoWillNowDie (approved):
epic!

PR #168, opened 2020-11-05
Run the movement calculations on the server

vainiovano:
This implementation has some problems that need to be fixed later:
* The clients receive a position update 60 times per second. The rate should be lower and the clients should interpolate between the states
* The clients send an action update 60 times per second. The rate should be lower, but implementing it properly can be difficult
* The player movements are very drifty on slow networks because there is no client-side prediction

video of laggy movement
laggy movement (100ms ping)


TheRyeGuyWhoWillNowDie (approved):
seems a TAD laggy, but can be fixed.

PR #169, opened 2020-11-06
Added spawn point system

### LINKED ISSUE #161: Player entry point entity
Create a new type of Node that can be placed into a scene. For the purposes of this spec, we will call it 'Player Spawn'. Upon entering the Main Game mode, the server will assign each player to an unoccupied Player Spawn, at random. That location will be used for the initial position of their character.

vincens2005:
Players line up around the spawn point.
When creating a map, make a position2d named "spawn" at the root of the scene and place it where you'd like players to start.


src/assets/main/main.gd
```diff
@@ -118,3 +119,14 @@ puppet func update_positions(positions_dict):
...
 	if not get_tree().is_network_server():
 		rpc_id(1, "player_moved", movement)
+
+
+func _on_maps_spawn(position):
+	# move players to spawn point
+	spawn_pos = position
+	var arrpos = 0
+	for i in players.keys().size():
+		if not players[players.keys()[i]].spawned:
```
TheSecondReal0: If spawned is true (which it is 100% of the time after the game starts) the players will not be moved to the spawn points when the map changes/when the round resets, is this a feature or a bug?

TheSecondReal0 (changes requested):

TheSecondReal0 (approved):

src/assets/main/main.gd
```diff
@@ -118,3 +121,19 @@ puppet func update_positions(positions_dict):
...
+	spawn_pos = position
+	recentmap = frommap
+	if frommap != "lobby":
+		notlobby = true
+	var arrpos = 0
+	for i in players.keys().size():
+		if notlobby and frommap == "lobby":
+			players[players.keys()[i]].spawned = []
+		if not frommap in players[players.keys()[i]].spawned:
+			players[players.keys()[i]].move_to(Vector2(position.x+((arrpos)*80),position.y),5)
```
vainiovano, line 137: `5` is not a valid vector. It prints errors to the console. You could use `Vector2(0, 0)`.

src/assets/main/main.gd
```diff
@@ -118,3 +121,19 @@ puppet func update_positions(positions_dict):
...
+			players[players.keys()[i]].spawned.append(frommap)
+		arrpos += 1
```
vainiovano, line 139: Using both `for i in players.keys().size()` and `arrpos` is redundant.

src/assets/main/main.gd
```diff
@@ -93,6 +95,7 @@ puppet func createPlayer(id, playerName):
 	$players.add_child(newPlayer)
 	newPlayer.setName(playerName)
 	print("New player: ", id)
+	_on_maps_spawn(spawn_pos, recentmap)
```
vainiovano, line 98: Spawning in the lobby should be a different code path because it needs collision detection with other players. Handling it in the same code path creates more complexity (`spawned` array). It can be this way for now until somebody knows what's the best way to do the collision detection (put a collider on each spawn?).

src/assets/maps/lobby/lobby.tscn
```diff
@@ -6,10 +6,10 @@
...
 
 [node name="ColorRect" type="ColorRect" parent="."]
-margin_left = -2048.0
-margin_top = -2048.0
-margin_right = 2048.0
-margin_bottom = 2048.0
+margin_left = -2174.0
+margin_top = -1927.0
+margin_right = 1922.0
+margin_bottom = 2169.0
```
vainiovano, line 12: probably an accidental change, isn't necessary

src/assets/main/main.gd
vainiovano: And why does `_on_maps_spawn` run on the clients? It should be only server-side

vainiovano:
The players can activate the clock task before the server moves them to the spawn.

PR #170, opened 2020-11-06
Clear out the 50+ warnings that are always present

TheSecondReal0:


vainiovano:
Why are some unused argument warnings disabled with `warning-ignore:unused_argument` and some with an underscore prefix? The Godot editor recommends to prefix unused arguments with an underscore.

TheSecondReal0:
> Why are some unused argument warnings disabled with `warning-ignore:unused_argument` and some with an underscore prefix? The Godot editor recommends to prefix unused arguments with an underscore.

I underscored unused vars in _process functions, everywhere else I put `warning-ignore:unused_argument` for consistency (don't want some state_changed functions to have underscored vars and some non-underscored vars, etc.). Both underscoring and adding a warning-ignore do the same thing, they just look different in the editor.

Another reason is so that people who have been working in their own scripts aren't disrupted by changing variable names.

GrbavaCigla (approved):

PR #172, opened 2020-11-09
Cleaning up a bit

GrbavaCigla:
I started working on modding system and I wanted to clean unnecessary code.
- Removed Control node with missing child
- Removed language code from mainmenu.gd
- Removed languages.gd since it is moved to settings


nicemicro (dismissed):
The language selection seems to work, however, in the "Settings" menu, only the first setting element changes according to the language.
Could you please check what's wrong and fix that, too in this PR?

nicemicro (approved):
The things removed are things that were no longer in use.
No bugs are added by this PR.

PR #173, opened 2020-11-09
Fixed back button in play menu

nicemicro:
Fixed the signaling problem that lead to a non-working "back" button in the play menu.


GrbavaCigla (approved):

vainiovano:
Why did PR #170 remove these lines? That PR has caused multiple regressions :(

nicemicro:
> Why did PR #170 remove these lines? That PR has caused multiple regressions :(

I have no idea. Our PR review process should be improved.

PR #174, opened 2020-11-09
Fixed network desync

TheSecondReal0:


src/assets/autoload/gamemanager.gd
```diff
@@ -24,7 +26,7 @@ func transition(new_state) -> bool:
 		var old_state: int = state
 		state = new_state
 		emit_signal('state_changed', old_state, new_state)
-		if get_tree().is_network_server():
+		if get_tree().is_network_server() and true:
```
TheSecondReal0: and true is there because without it github doesn't register the change to that line

vainiovano (approved):
Works well, fixes the problems

PR #175, opened 2020-11-09
Interactarea fix

TheSecondReal0:
Fixes the bug where all players get the UI popup when one comes into range with an interactpoint.


nicemicro (approved):
It does what it says it does.

PR #177, opened 2020-11-10
Runtime player spawning and respawning

TheSecondReal0:
All players are now created at runtime, including the host. All player nodes are now deleted and recreated upon map switch.

### DIFF
--- a/src/assets/autoload/network.gd
+++ b/src/assets/autoload/network.gd
@@ -78,7 +78,10 @@ func _player_connected(id) -> void:
 
 func _player_disconnected(id) -> void:
 	peers.erase(id)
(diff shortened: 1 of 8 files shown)


vainiovano (dismissed):
Looks great!

Some notes:
- Spawning new players to the lobby uses only one spawn point. A better collision-based system can be implemented in the future
- The old spawning code and `spawned` array could be removed

nicemicro:
>     * The old spawning code and `spawned` array could be removed

I'd suggest this being fixed before the PR is merged.

vainiovano:
The players still have a `spawned` array that's unused

vainiovano (approved):
Looks good, works well

PR #180, opened 2020-11-10
Multiple spawnpoints

TheSecondReal0:
There can now be multiple spawnpoints. Spawnpoints are used in the order they load in the scene tree, so the first player will spawn at the topmost position2D under the spawnpoints node inside the map scene.


vincens2005 (commented):
It all looks good to me, but con someone test it? I can't test it right now.

nicemicro (approved):
Seems good to me.

PR #187, opened 2020-11-12
Implement simple client-side prediction

vainiovano:
The client stores the unreceived inputs in a queue and executes them on top of the server's player position. Current problems:
- The players can jump a bit because of a mismatch between local state and server state. It can occur for example because of unstable ping. The position update rate should be lower to allow implementing soft interpolation
- Client-side prediction doesn't work correctly if _physics_process isn't called at a constant rate.

You can add a delay on loopback/localhost on Linux by running `tc qdisc add dev lo root netem delay 100ms` as root (run `tc qdisc del dev lo root netem delay 100ms` to remove the delay).

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -8,10 +8,13 @@ var player_scene = load(player_s)
 var players = {}
 #!!!THIS IS IMPORTANT!!!
 #INCREASE THIS VARIABLE BY ONE EVERY COMMIT TO PREVENT OLD CLIENTS FROM TRYING TO CONNECT TO SERVERS!!!
-var version = 12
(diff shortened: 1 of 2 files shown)


nicemicro:
When I crank up the latency to 500 ms, the client jitters noticeably after the movement stopped, first it seemed to be dependent on how long was I moving before stopping, but it's quite random..

vainiovano:
I don't think that I can fix that. The server should send position update packets at a lower rate and the client should slowly correct the main player's position. That would make movement more soft.

nicemicro (approved):
I approve it for merge, as it is a step in the right direction, and it can be improved later.

PR #189, opened 2020-11-12
changed alien to spawnpad

jgx0:
I changed the alien to a spawnpad. It looks much better


TheSecondReal0 (approved):
Works perfectly, looks good!

PR #181, opened 2020-11-10
Basic test map

TheSecondReal0:
Added rooms to the test map, I also moved some of the demos around.

PR #191, opened 2020-11-13
Clean up merge of #180 and #181

TheSecondReal0:


nicemicro (approved):
It does what it says it does without breaking anyithing, good to go.

PR #200, opened 2020-11-16
UI classes

TheSecondReal0:
Creates UI classes for future menus and tasks to extend. 

Improvements:
* Less arbitrary work when creating UI. Without this, whenever you make a UI you have to include a bunch of system interaction, which bogs down actual development
* No longer need to include open() and close() in UI scripts, closing can be done by calling hide() and the classes will handle everything else
* Flexible, if you want to override low level operation you can by defining base_close() or base_open() in your UI script

Standard operation:
1. UIManager tells uicontroller.gd to open a menu
2. uicontroller.gd calls base_open() on the menu, base_open() handles all of the lower level stuff like communicating with UIManager
3. uicontroller.gd calls open() on the menu if it has that method. This means that if you want your menu to do something when it opens you can, otherwise you don't need to think about it
4. UIManager tells uicontroller.gd to close the menu
5. uicontroller.gd calls base_close() on the menu, again handles lower level stuff like communicating with UIManager
6. uicontroller.gd calls close() on menu if it has that method. This means that if you want your menu to do something when it closes you can, otherwise you don't need to think about it

For developers:
* If you're making a menu, make sure your script extends one of the base classes (for ex. WindowDialog -> WindowDialogBase)
* If you're making a task, make sure your script extends one of the task classes (for ex. WindowDialog -> WindowDialogTask)
* You must set the menu name in the inspector for your menu to mesh with the UI manager, along with adding it to to the list of menus in uimanager.gd

Notes:
* All task classes extend their respective base classes. PopupTask extends PopupBase, WindowDialogTask extends WindowDialogBase, etc.
* There are multiple UI classes so that the individual behavior of each UI node isn't lost when one of these classes is used
* Eventually task classes will handle communication with the task manager, but a task manager does not exist so that has to wait


TheRyeGuyWhoWillNowDie (approved):
epic!

nicemicro (approved):
Looks fine to me, and easily extendable.

PR #205, opened 2020-11-21
Player can pick up or drop items

zapzoop0099:
This is a very very basic implementation of picking or droping items..

Use controls p for pick and q for drop
A lot of work is still needed to done please feel free to comment this or in matrix and suggested the detail which should be added

To test this:- chainlink the item scene(res://assets/maps/common/item/item.tscn) into test map.


moxvallix:
Just a suggestion: might be best to bind pickup and drop to the same key. Nicer UX.

zapzoop0099:
This becomes outdated!!

PR #209, opened 2020-11-23
UI and map interact resources

TheSecondReal0:
Adds resources that can be customized in the editor to open UIs or to handle map interaction. The resource types are added via a plugin.

I've updated interactpoint and standbutton to use resources instead of an assortment of exported vars, go look at them to see an example of how to use the resources.

### HOW TO USE
In any node of your choice:
1. Add an export var of type resource with ```export(Resource) var my_var``` or ```export var my_var: Resource```
2. In the editor inspector, set that variable to an Interact resource
3. Set "Interact Type" to the type of interaction you want
4. Customize the resource corresponding to that type
5. When you want the resource to execute the interaction, call interact(self) on it with ```my_var.interact(self)```

### NOTES
* You MUST pass self in interact(), otherwise there is no way to reference the node you want to interact with. This is due to a technical limitation with the resource class, and cannot be fixed/removed. 
* For the same reason, you should pass self in get_interact_data(), although this is only required the first time you call get_interact_data() or before you call interact() for the first time
* Because the resources have tool scripts attached, the resource name cannot be modified. I will look into fixing this at a later date.

This PR doesn't change the functionality of any systems, it only provides a wrapper to utilize them. This should not cause any loss of functionality for scripts that interact with systems and managers directly.

### HOW IT WORKS
This PR creates a plugin that adds custom resource types, listed below:
#### InteractUI:
Handles opening UIs. It does this by taking the values set in the editor and translating them into a call to the UIManager.

This resource overrides get(), set(), and get_property_list() to add a variable group to the inspector named "advanced." Currently, this group only contains the reinstance bool, which tells the UI system whether or not to delete and re-instance UI nodes when they are opened. This resource forces itself to be local to scene.

#### InteractMap:
Handles interacting with map nodes. It translates values set in the editor into a call to the MapManager, similar to InteractUI. 

This resource forces itself to be local to scene.

#### Interact:
Acts as a wrapper for InteractUI, InteractMap, and future interact types. When interact() is called, the resource will just forward it to the correct nested resource. When get_interact_data() is called, it calls get_interact_data() on the correct nested resource and adds the corresponding interact type (ui = 1, map = 2) if the return dictionary doesn't already contain it. 

This resource overrides get(), set(), and get_property_list() to restrict ui_resource to the InteractUI resource type and map_resource to the InteractMap resource type. This resource forces itself to be local to scene.

Using this resource is not required, you can directly use InteractUI and InteractMap as well. 

### WHY TO USE IT
There are two main reasons to use these resources instead of interacting with the systems directly:
1. Ease of use/accessibility. Using these resources negates the need to understand how the core systems work, which makes development easier and faster.
2. Reduces impact of a system rewrite. Since the resources handle all system interaction, only the resource scripts would have to be adapted.

### PLUGIN INFO
* Plugin file is res://addons/opensusinteraction
* Plugin script is res://addons/opensusinteraction/opensusinteraction.gd, this is what adds the custom resource types to the editor
* Each resource type is stored in a folder named after their class, ex. InteractUI files are stored in res://addons/opensusinteraction/resources/InteractUI
* There is a blank version of each resource in their respective folders. These can be loaded anywhere without worry as each resource will automatically set themselves local to scene, so they will never be modified unless directly opened
* I plan to extend this plugin to have a Task resource type, and to add custom editors if necessary


nicemicro (commented):
As far as I can see, it doesn't break existing functionality.
Have not tested the stuff it brings to the project in terms of simplifying extensibility, I will wait one more day for others to review, and if no one will object, I'll approve the PR.
If anyone else does a more thorough testing, feel free to approve it immediately.

TheRyeGuyWhoWillNowDie (approved):
yes

vincens2005 (approved):

PR #215, opened 2020-11-24
Including the customizable vector art character

nicemicro:
At this stage this will break the game, I put it here so you can see and comment on how the implementation is going.


nicemicro:
This has been picked up and extended by @TheRyeGuyWhoWillNowDie, so this PR is not useful anymore.

PR #214, opened 2020-11-24
Map is now syncd between server and client, custom maps are now possible.

TheRyeGuyWhoWillNowDie:
Alright so THIS WILL CRASH UNLESS: take test.tscn from maps and put it into /home/YOUR HOME/.local/share/opensuspect/maps/servermap.tscn

PR #217, opened 2020-11-24
Task manager and task resource

### LINKED ISSUE #103: Task management system
A system to manage what needs to be done before a task is completed and what happens after it is completed. Features to include:
- Keep track of which tasks are accessible to the player (based on prerequisite tasks or player location)
- Prerequisites (allow for tasks require certain conditions to be met before they can be completed. For instance, a task to open a door which requires you to complete another task where you grab a key)
- Effects (opening doors, unlocking another task, etc.. Will need knowledge of the current map for dynamic map objects like doors.)

TheSecondReal0:
The task system has basic functionality and has groundwork laid for future development. The InteractTask resource is used the same way as other interact resources, see #209. The biggest change coming in this PR is that init_resource(self) should be called under _ready() of its attached node. This allows the task resource to notify the task manager of its existence when the map is loaded.

It is up to reviewers if this should be merged right now or if it should wait until it is more fully featured.

#### HOW IT WORKS
The resource will take what's been set in the editor and translate it into information to give to the task manager. The task manager will then process this information and register a new task under a unique task ID. 

Currently, nothing is done with tasks but groundwork is laid to advance task states. This doesn't happen yet because it requires task UI to integrate with the task system as well, which is beyond the scope of this PR.

#### WHAT'S LEFT
* Task assignment
* Task prerequisites (items, conditions, etc.)
* Task outputs (items, map interactions, other tasks, etc.)
* Two way integration. Right now only the resource integrates with the task manager, in the future task UIs must do this as well. I plan to set it up so that task UIs integrate with the task resource, which then integrates with the task manager.

NOTE: The game crashes when you attempt to transition the game to the test map, this is because of #214 and NOT this PR. See #220 for the fix.


nicemicro (commented):
I have tested it with rebasing it on top of PR-220 (which I have just approved for merge), and while I can't comment on code quality due to my lack of proper understanding of the resource system, I have checked and things work.
One question I have is:
what is the expected behavior when we are on top of the clockset task that doesn't use the task system?

TheSecondReal0:
> One question I have is:
> what is the expected behavior when we are on top of the clockset task that doesn't use the task system?

It should behave exactly as it did before, I haven't changed anything about that. I haven't switched it yet because the task system doesn't support the behavior of the clockset task. Soon I'll add support for the task UI to send data through the task resource into a map interaction upon completion, which is when the standbutton clockset task can be migrated.

nicemicro:
> It should behave exactly as it did before, 

All right, so sometimes it gets randomly assigned to someone and sometimes it doesn't, and in the first case it works as it used to, on the latter, there's no interaction.
So everything works as intended.

nicemicro (dismissed):

TheSecondReal0:
All meaningful commits included in #222, closing

PR #220, opened 2020-11-28
Revert #214, fix crashes

TheSecondReal0:
Completely reverts #214

PR #222, opened 2020-11-29
Rebased verion of #217

nicemicro:
I have rebased at home, and it doesn't have any merge conflicts.
And it seems to work fine for me.

EDITING TO INCLUDE INFO FROM ORIGINAL PR - TheSecondReal0

The task system has basic functionality and has groundwork laid for future development. The InteractTask resource is used the same way as other interact resources, see #209. The biggest change coming in this PR is that init_resource(self) should be called under _ready() of its attached node. This allows the task resource to notify the task manager of its existence when the map is loaded.

It is up to reviewers if this should be merged right now or if it should wait until it is more fully featured.

#### HOW IT WORKS
The resource will take what's been set in the editor and translate it into information to give to the task manager. The task manager will then process this information and register a new task under a unique task ID. 

Currently, nothing is done with tasks but groundwork is laid to advance task states. This doesn't happen yet because it requires task UI to integrate with the task system as well, which is beyond the scope of this PR.

#### WHAT'S LEFT
* Task assignment
* Task prerequisites (items, conditions, etc.)
* Task outputs (items, map interactions, other tasks, etc.)
* Two way integration. Right now only the resource integrates with the task manager, in the future task UIs must do this as well. I plan to set it up so that task UIs integrate with the task resource, which then integrates with the task manager.

NOTE: The game crashes when you attempt to transition the game to the test map, this is because of #214 and NOT this PR. See #220 for the fix.


TheSecondReal0 (approved):
All functionality is there and merge conflicts are fixed, thanks @nicemicro!

PR #223, opened 2020-11-30
Switched to Skeleton2D animation for more flexibility in animations

jngo102:
Switched to `Skeleton2D` animation to allow for a greater range of keyframe poses.
- `Node2D` named Skeleton contains the polygons with textures and the actual `Skeleton2D` node that controls those polygons.
- `$spritecollection/AnimationPlayer` animates this Skeleton.
- Logic for flipping Skeleton container in player.gd when turning around.
- `$spritecollection` Sprites are kept in case we want to revert back for any reason.


TheRyeGuyWhoWillNowDie (approved):
good

PR #225, opened 2020-12-01
Added scrollwheel zoom

vincens2005:
There's currently no solution for mobile, but if you scroll while in a scene with the player, your view will zoom in


TheRyeGuyWhoWillNowDie (approved):
yes

PR #229, opened 2020-12-02
Killing and reloading (#221)

### LINKED ISSUE #221: Killing and reloading
We need to revamp the killing system. What should be done:

1. an overlay for the target to kill (crosshairs maybe?),
2. a customizable keystroke + an on-screen clickable icon to commit the kill and
3. a customizable keystroke + an on-screen clickable icon to reload the gun which also might serve as the timer for the reload.

For details, see [the relevant part of the design document](https://github.com/opensuspect/design-document/blob/main/general_gameplay.md#kill-mechanics).

jngo102:
- Re-implemented red kill outline when near other players as an infiltrator
- Replaced placeholder kill `Sprite` with custom kill and reload `TextureButtons`
- Added `Node2D` named `DeathHandler` which handles player deaths
- Migrated animations to use an `AnimationTree` for more powerful animation blending
- Execute and sync deaths across the network


src/assets/autoload/network.gd
```diff
@@ -171,6 +171,16 @@ func get_player_name(id: int = myID) -> String:
 func get_peers() -> Array:
 	return peers
 
+func get_rpc_id_from_player_name(player_name: String) -> int:
```
TheSecondReal0: I don't know if there's a specific reason for crawling the scene tree, but using the ```names``` dict in network.gd would be much easier and cleaner.

Additionally, in other areas of the game RPC IDs are referred to as network IDs. Please change this function name for the sake of consistency.

src/assets/main/main.gd
```diff
@@ -141,3 +143,35 @@ master func _on_maps_spawn(spawnPositions: Array):
...
+		infiltrator_killed_player(killer_id, killed_player_id)
+	else:
+		rpc_id(1, "infiltrator_killed_player", killer_id, killed_player_id)
+
+remote func infiltrator_killed_player(killer_id: int, killed_player_id: int) -> void:
+	"""
+	Runs on the server; sends an RPC to every player to indicate that a
+	particular player has been killed.
+	"""
+	for player_id in players.keys():
```
TheSecondReal0, line 173: Could add a check here just to make sure the function is running on the server, something like this:
```gdscript
        return
```

src/assets/main/main.gd
```diff
@@ -141,3 +143,35 @@ master func _on_maps_spawn(spawnPositions: Array):
...
+		if players[player_id].main_player:
+			# Can't RPC on self
+			player_killed(killer_id, killed_player_id)
+		else:
+			rpc_id(player_id, "player_killed", killer_id, killed_player_id)
+
+remote func player_killed(killer_id: int, killed_player_id: int) -> void:
```
TheSecondReal0: Switch ```remote``` for ```puppet```, this guarantees the function can only be called remotely by the server

src/assets/player/death_handler.gd
```diff
@@ -0,0 +1,19 @@
...
+onready var player: KinematicBody2D = get_owner()
+onready var main: YSort = get_tree().get_root().find_node("Main", true, false)
+
+# Whether the player is dead
+var is_dead: bool = false
+
+func die_by(killer_id: int) -> void:
+	"""Player death."""
+	is_dead = true
+	var killer: KinematicBody2D = main.players[killer_id]
```
TheSecondReal0: There's probably a cleaner way to get the list of player nodes, perhaps by storing them in PlayerManager when they are created. This is not a necessary change for merging, but highly recommended.

src/assets/player/infiltrator.gd
```diff
@@ -90,56 +92,54 @@ func _get_target() -> void:
...
 
 func _instantiate_kill_gui() -> void:
-	"""
-	Add kill UI to the infiltrator's HUD
-	"""
-	ui_controller.open_menu("killui", {"linked_node": self, "rect_position": Vector2(850, 500)}, true)
+	"""Add kill UI to the infiltrator's HUD."""
+	var viewport_size : Vector2 = get_viewport().get_visible_rect().size
+	var kill_ui_position := Vector2(viewport_size.x * 0.75, viewport_size.y * 0.75)
+	ui_controller.open_menu("killui", {"linked_node": self, "rect_position": kill_ui_position}, true)
```
TheSecondReal0: It seems that you're trying to put the buttons in a set place in the corner of the screen, I think using anchors in the killui scene would be a better solution. Setting all margins to 0 with the top and left anchors to .75 along with bottom and right anchors to 1 should work.

src/assets/player/viewport_texture_target.gd
```diff
@@ -0,0 +1,5 @@
+extends Sprite
+
+func _ready() -> void:
+	# Duplicate the ShaderMaterial so that changes to shader parameters in one player material don't affect all players
+	set_material(material.duplicate())
```
TheSecondReal0: You might be encountering a weird bug, but I think this is what the ```local_to_scene``` resource setting is for.

TheSecondReal0 (changes requested):
I've made a few review comments, they're mostly trying to keep the codebase as clean as possible. None of them really effect functionality. I tested this in editor and exported, everything works perfectly!

One thing that isn't in the comments is that I would like you to organize the texture files inside ```assets/ui/hud/infiltrator_hud```. This basically means putting them into a folder named "textures" or something inside ```assets/ui/hud/infiltrator_hud```.

All in all, this is a great PR and a giant leap towards alpha!

src/assets/autoload/network.gd
jngo102: I should've been more clear about this method's description, but it actually gets the network ID of the player from the name of the player node on the client side, i.e. the KinematicBody2D, not the player's entered name.

TheSecondReal0: OK, looking at the code it looks like the name stored in network.gd should be the same as the name of the player node, but this is more flexible and would survive if we changed that later on. 

One thing I would suggest is moving this out of network.gd, since it doesn't really belong there. I'm not quite sure where to put it if you plan on using this function from a lot of places, but it could just be in the script that calls it. It's your choice to make this change.

TheSecondReal0 (dismissed):
Everything looks good! I've replied to one of the comments you made about moving get_network_id_from_player_node_name() to somewhere else. Unless you plan on using this function from a lot of places, I think you could move it into main.gd. It's completely up to you to merge it how it is or to move the function.

TheSecondReal0 (approved):
Tested everything, code is clean, overall great PR, approved!

Issue #211, opened 2020-11-23
Dying


nicemicro:
Players who are killed should die and became a spectator.

PR #231, opened 2020-12-03
Spectator mode

jngo102:
- Create a corpse upon death and turn the player into a ghost that can only be seen by other ghosts
- Added resurrect animation that plays after death and while becoming a ghost
(#211)


TheRyeGuyWhoWillNowDie (approved):
very yes

PR #232, opened 2020-12-03
added win conditions

TheRyeGuyWhoWillNowDie:
that's it.

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -171,6 +171,7 @@ remote func infiltrator_killed_player(killer_id: int, killed_player_id: int) -> 
 		return
 
 	for player_id in players.keys():
+		PlayerManager.playerRoles.erase(killed_player_id)
 		if players[player_id].main_player:
 			# Can't RPC on self
 			player_killed(killer_id, killed_player_id)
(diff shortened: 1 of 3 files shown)


src/assets/main/main.gd
```diff
@@ -179,8 +180,25 @@ remote func infiltrator_killed_player(killer_id: int, killed_player_id: int) ->
 
 puppet func player_killed(killer_id: int, killed_player_id: int) -> void:
 	"""Runs on a client; responsible for actually killing off a player."""
-	var killed_player_death_handler: Node2D = players[killed_player_id].get_node("DeathHandler")
-	killed_player_death_handler.die_by(killer_id)
+	PlayerManager.playerRoles.erase(killed_player_id)
```
TheSecondReal0, line 183: DO NOT ERASE ROLES

We should keep track of them so they can be displayed later (must keep track of who the impostors are to show them on a win screen or something)

src/assets/main/main.gd
```diff
@@ -179,8 +180,25 @@ remote func infiltrator_killed_player(killer_id: int, killed_player_id: int) ->
...
+	var killvalid = false
```
TheSecondReal0, line 184: Give this var an accurate name, it seems to keep track of whether or not the round will end instead of the validity of the kill. 

You could also rename it to be camelCase or snake_case, along with other variables

TheSecondReal0 (changes requested):
Please avoid destroying data that we will realistically want to use later. There are a few other readability concerns with this PR as well.

Otherwise, this PR adds some crucial functionality and works perfectly.

TheRyeGuyWhoWillNowDie:
maybe someone else can do it better

Issue #227, opened 2020-12-01
UI exits game, takes back to menu screen


moxvallix:
Woo, this is my first bug report!

Description of Bug:
Pressing exit from the keybinds menu whilst in a game, takes the player back to the menu screen, rather than back to the game.

Steps to reproduce:
1. Host a game
2. Open the menu, go Settings > Keybinds
3. Press the back button

TheRyeGuyWhoWillNowDie:
this was fixed

PR #234, opened 2020-12-03
UI exits game, takes back to menu screen

jngo102:
- Fixed pressing Back or Save button in keybinds screen returning to menu
(#227)


TheSecondReal0 (approved):
While not perfect, it's the best that could be done without completely redesigning the keybinds menu. Tested and everything works perfectly.

PR #237, opened 2020-12-04
Clean up UI system

TheSecondReal0:
* Added a way to ignore nodes instanced via the editor in update_instanced_uis(). This drastically reduces the amount of errors pushed.
* Infiltrator hud no longer blocks interactui buttons
* Lobby chat button no longer crashes game


vincens2005 (commented):
Seems good, but I haven't tested it. Will test tomorrow morning if no one else has

TheRyeGuyWhoWillNowDie (approved):

PR #239, opened 2020-12-04
Added loading custom content from PCK files

TheRyeGuyWhoWillNowDie:
added button on menu that loads custom.pck in .local/share/opensuspect


jngo102 (approved):
Works well enough.

PR #241, opened 2020-12-05
elimination victory

TheRyeGuyWhoWillNowDie:
that's it. x2

### DIFF
--- a/src/assets/autoload/playermanager.gd
+++ b/src/assets/autoload/playermanager.gd
@@ -9,6 +9,7 @@ var ournumber
 var tasks = [-1]
 var taskstoassign
 var assignedtasks
+var aliveplayers: Dictionary = {}
 #vars for role assignment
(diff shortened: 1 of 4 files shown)


nicemicro (dismissed):
This PR isn't really focused on win conditions, rather on counting the remaining players, which is a relevant thing to do on the first map, but if we are planning to have a no-kill map in the future (which is one of the suggestions for the future), than it will have to be totally rewritten.

My suggestions:

- have a separate function to enumerate the players in each team
- have a function that checks which team has more players
- have a separate function that triggers the transition to the lobby (which can be later extended with the win / loss announcement screens)
- have a separate, dedicated function that can be used to assess the win conditions (in this case, only the number of players)
- use the player_killed function to call the win condition assessment function (later other functions like tasks on completion could call the win condition assessment function too)

PR #245, opened 2020-12-06
Win condition plumbing + majority victory win condition implemented

### LINKED ISSUE #195: Win conditions
Handling for the game to switch to the win/lose screen.

Should have:
- A system to handle conditions based on other in-game variables
- System to determine winners / losers, and triggers end screen accordingly

Conditions:
- Infiltrators win when number of infiltrators >= number of agents
- Agents win when tasks are complete
- Agents win when there are no infiltrators

nicemicro:
Based on @TheRyeGuyWhoWillNowDie 's original idea, I have separated functionality related to win conditions into separate functions.
The elimination victory / gaining majority victory condition has been implemented in this framework in a flexible way.
It currently doesn't include victory scenes, only instant transition back to lobby.

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -41,6 +41,14 @@ func _physics_process(_delta: float) -> void:
 			if id != 1:
 				rpc_id(id, "update_positions", positions_dict, players[id].input_number)
 
+func main_player_id():
+	"""Returns the id of the main player (the player who is the playable character
+	on this instance)"""
+	for player_id in players.keys():
+		if players[player_id].main_player:
+			return player_id
+	return -1
+
 func connection_handled(id: int, playerName: String) -> void:
 	print("connection handled, id: ", id, " name: ", playerName)
 	if not get_tree().is_network_server():
@@ -192,8 +200,16 @@ puppet func end_round(winner):
 	"""This function is called by the server and when it is, it would need to
 	show the win / lose screens for the players, and then transitions back
 	to the lobby."""
-	#Here should be the code for displaying the victory / defeat screens
 	#print("Round ended with winning team ", winner)
+	#print("Current player's team: ", PlayerManager.get_player_team(main_player_id()))
+	if PlayerManager.get_player_team(main_player_id()):
+		#Here should be the code for displaying the victory screen
+		#print("We have won the game")
+		pass
+	else:
+		#Here should be the code for displaying the defeat screen
+		#print("We have lost the game")
+		pass
 	GameManager.transition(GameManager.State.Lobby)
 
 master func victory_check():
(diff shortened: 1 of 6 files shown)


TheSecondReal0 (changes requested):
There's a bug when the elimination victory is triggered, the last person who's killed is still dead/dying on the impostor client and the murdered client in the lobby. It works perfectly for every other client though.

nicemicro:
> There's a bug when the elimination victory is triggered, the last person who's killed is still dead/dying on the impostor client and the murdered client in the lobby. It works perfectly for every other client though.

That is true, I have no idea why though, so I request assistance..

nicemicro:
Thanks for pointing me in the good direction @TheSecondReal0 and for @zapzoop0099  for the help with the networking code.
The bug does not exist anymore, as the checks have been moved to a better place.
Ready for re-review.

src/assets/main/main.gd
```diff
@@ -41,6 +41,14 @@ func _physics_process(_delta: float) -> void:
...
 				rpc_id(id, "update_positions", positions_dict, players[id].input_number)
 
+func main_player_id():
+	"""Returns the id of the main player (the player who is the playable character
+	on this instance)"""
+	for player_id in players.keys():
+		if players[player_id].main_player:
+			return player_id
+	return -1
+
```
vainiovano, line 52: Is the returned id always equal to `Network.myID` or am I missing something?

src/assets/main/main.gd
```diff
@@ -182,6 +196,81 @@ puppet func player_killed(killer_id: int, killed_player_id: int) -> void:
...
+	#TODO:
+	#Here, there should be a check whether the current map allows for elimination victory ot not
+	if victorious == -1:
+		victorious = elimination_victory_check(0)
+	#TODO
+	#Here, all other victory conditions should be checked.
+	if victorious != -1:
+		for player_id in players.keys():
+			if players[player_id].main_player:
+				# Can't RPC on self
```
vainiovano: RPCing self is possible when using `sync` versions of functions (`remotesync`, `mastersync`, `puppetsync`) (see https://github.com/godotengine/godot/pull/28783)

src/assets/main/main.gd
nicemicro: I've seen the main player object being returned similarly in `playermanager.gd` 's `get_main_player()` function.
The only time it wouldn't equal to `Network.myID` or equally the return value of `Network.get_my_id()` would be when run on a dedicated server with no main player.
But as I see I haven't checked for that possibility on the victory / defeat screen selection, so I'm gonna do that.

src/assets/main/main.gd
nicemicro: Fixing, and also doing the same for `player_killed` function

jngo102 (approved):
Encountered no issues during testing.

TheSecondReal0 (approved):

PR #246, opened 2020-12-06
Items and inventory (#194)

### LINKED ISSUE #194: Items and inventory

Should have:
- One item at a time

jngo102:
- Implemented picking up and dropping items, synced over the network
- Modified MapManager and base map scene


src/assets/player/item_handler.gd
```diff
@@ -0,0 +1,72 @@
...
+func drop(item: KinematicBody2D) -> void:
+	"""Drop an item."""
+	var map_items: Node2D = MapManager.get_current_map().items
+	var offset: Vector2 = player.global_position - map_items.global_position
+	remove_child(item)
+	map_items.add_child(item)
+	item.position = offset
+	item.set_collision_layer_bit(4, true)
+	if player.has_node("Infiltrator"):
+		player.get_node("Infiltrator").enable_killing(true)
```
vainiovano: this doesn't work when killing a player that doesn't hold an item (the item is Nil)

src/assets/main/main.gd
```diff
@@ -148,6 +151,44 @@ func _on_main_player_moved(movement: Vector2, velocity: Vector2, last_input: int
 	if not get_tree().is_network_server():
 		rpc_id(1, "player_moved", movement, velocity, last_input)
 
+func _on_main_player_picked_up_item(item_path: String) -> void:
+	if get_tree().is_network_server():
+		player_picked_up_item(item_path, 1)
+	else:
+		rpc_id(1, "player_picked_up_item", item_path, Network.get_my_id())
```
vainiovano: using `remotesync` and just rpcing the server could also work

src/assets/main/main.gd
```diff
@@ -148,6 +151,44 @@ func _on_main_player_moved(movement: Vector2, velocity: Vector2, last_input: int
...
+
+func _on_main_player_dropped_item() -> void:
+	if get_tree().is_network_server():
+		player_dropped_item(1)
+	else:
+		rpc_id(1, "player_dropped_item", Network.get_my_id())
+
+remote func player_picked_up_item(item_path: String, id: int) -> void:
```
vainiovano: you could probably use `get_rpc_sender_id` for these functions if they were always called using rpcs

src/assets/main/maps.gd
```diff
@@ -1,42 +1,46 @@
...
 		GameManager.State.Normal:
-			switchMap('test')
+			switch_map("test")
 
-func switchMap(newMap: String) -> void:
-	print('switchMap called for ', newMap)
-	if not maps.keys().has(newMap):
+func switch_map(new_map_name: String) -> void:
+	print("switch_map called for ", new_map_name)
+	var map_path: String = Helpers.find_file(new_map_name + ".tscn", maps_path)
```
vainiovano, line 25: map file names could also contain a hash of the map to prevent name collisions

vainiovano (commented):
Handling item position syncing like this can result in inexact item positions if the server sends player position updates at a lower rate.

src/assets/player/item_handler.gd
jngo102: Added null check

src/assets/main/main.gd
jngo102: Replaced `Network.get_my_id()` with `get_rpc_sender_id()`

src/assets/main/main.gd
jngo102: Changed to `remotesync`

TheRyeGuyWhoWillNowDie (approved):
good

src/assets/main/maps.gd
```diff
@@ -1,42 +1,47 @@
...
+	var path_checksum: String = map_path.sha256_text()
+	if map_path == "":
 		return
-	print("loading map: ", newMap)
+	var map_scene: PackedScene = load(map_path)
+	print("loading map: ", new_map_name)
```
TheSecondReal0, line 30: It should print "loading map" before the map is loaded, that way it's more useful for debugging.

TheSecondReal0 (changes requested):
The character is invisible in the lobby and the role screen crashes the game. It crashes on line 7 of ```item_handler.gd```, which is ```onready var item_pickup_range: Area2D = player.get_node("ItemPickupRange")```. The crash is unique to the role screen, as I can disable it and avoid the crash. The character is still invisible in the lobby. For some reason if I click continue through the error, the character is shown with the correct customization on the role screen. The crash repeatedly happens on each continue when the role screen should be shown.

The crash generally seems to be caused by the item handler not being able to grab the player node correctly. Switching ```get_owner()``` to ```get_parent()``` in ```onready var player: Player = get_parent()``` in ```item_handler.gd``` revealed that it's getting set to a Node2D instead of a Player class, this leads me to believe that the item handler script is being loaded somewhere incorrectly.

PR #247, opened 2020-12-06
Items pick and drop functionality

### LINKED ISSUE #194: Items and inventory

zapzoop0099:
This PR gives the player ability to pick and drop items 

**Features:**

- Player can pick items and drop it and this sync over network

- If the item is picked by an infiltrator killing is disabled

- Items will hover in air when not picked

**This pr doesn't include:-**

- Items related to tasks

- Yellow shader over the items when get close to it

- Any type of asset(It have just a small box as placeholder for item prop )

The map already contained some instanced scene of items so when it get reviewed i will remove them at last
It also doesn't any include good comments :( but if it stays here for long i will push them
Please also refer to pr #246  and merge the best one.

### DIFF
--- a/src/assets/main/main.gd
+++ b/src/assets/main/main.gd
@@ -15,6 +15,10 @@ var newnumber
 var spawn_pos = Vector2(0,0)
 var player_data_dict: Dictionary
 
+var hand_pos:Dictionary
+var pick_range:Dictionary
+var keystroke:int
+
 signal positions_updated(last_received_input)
 
 func _ready() -> void:
@@ -100,6 +104,8 @@ puppetsync func createPlayer(id: int, playerName: String, spawnPoint: Vector2 = 
 	var newPlayer = player_scene.instance()
 	newPlayer.id = id
 	newPlayer.setName(playerName)
+	hand_pos[id] = newPlayer.get_node("reach/hand_pos")
+	pick_range[id] = newPlayer.get_node("reach")
 	#newPlayer.set_network_master(id)
 	if id == Network.get_my_id():
 		newPlayer.main_player = true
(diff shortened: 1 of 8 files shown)


src/assets/main/main.gd
```diff
@@ -276,3 +282,62 @@ func _apply_customizations(player: KinematicBody2D, player_data: Dictionary) ->
 	face_wear.texture = load(appearance["Face Wear"]["texture_path"])
 	hat_hair.texture = load(appearance["Hat/Hair"]["texture_path"])
 	mouth.texture = load(appearance["Mouth"]["texture_path"])
+
+
+
+func _input(event):
+	if Input.is_action_just_pressed("ui_pick"):
```
jngo102, line 289: Putting an `Input.is_action_just_pressed` in `_input` doesn't make use of `event`.
Try using `event.is_action_pressed` instead.

src/assets/main/main.gd
```diff
@@ -15,6 +15,10 @@ var newnumber
 var spawn_pos = Vector2(0,0)
 var player_data_dict: Dictionary
 
+var hand_pos:Dictionary
+var pick_range:Dictionary
+var keystroke:int
```
jngo102, line 20: Is `keystroke` ever used? What's its purpose?

src/assets/main/main.gd
```diff
@@ -276,3 +282,62 @@ func _apply_customizations(player: KinematicBody2D, player_data: Dictionary) ->
...
+	
+	
+func get_hand_pos(id):
+	return hand_pos.get(id)
+	
+func get_pick_range(id):
+	return pick_range.get(id)
+
+
+func enable_killing(id):
```
jngo102, line 328: I'm not sure the expensive tree parsing in `enable_killing` and `disable_killing` are necessary here when you could just use
```
if players[id].has_node("Infiltrator"):
    players[id].get_node("Infiltrator").enable_killing([true/false])
```

jngo102 (changes requested):

zapzoop0099:
Didn't match with community standards

PR #250, opened 2020-12-07
Improved lobby scenery

jngo102:
- Random spawning of scenery objects
- Implemented object pooling in a class


src/assets/autoload/helpers.gd
```diff
@@ -1,5 +1,12 @@
 extends Node
 
+func choose(array: Array):
```
TheSecondReal0: Maybe change this function name to ```pick_random()```?

TheSecondReal0 (changes requested):
There's one small change to a function name I'd like you to make but other than that everything looks good! The code looks good and it works perfectly!

src/assets/autoload/helpers.gd
jngo102: Changed

TheSecondReal0 (approved):
Code is solid, works perfectly, ready to merge!

PR #251, opened 2020-12-07
Add a base UI class for control nodes

TheSecondReal0:
This creates a base class to use when creating UI elements with a control node as the root. It also adds a toggle for the UI to block movement.

Current UIs with a Control node as the root have been updated.


jngo102 (approved):
Nice rework of some of the UI. This will be useful.

PR #254, opened 2020-12-08
Items and inventory (#194)

jngo102:
- Pick up and drop items with "E" or by clicking on them
- Modified MapManager for easier access to items in maps
- Fixed bugs caused by merging updated main into branch


src/assets/main/main.gd
```diff
@@ -157,6 +160,46 @@ func _on_main_player_moved(movement: Vector2, velocity: Vector2, last_input: int
 	if not get_tree().is_network_server():
 		rpc_id(1, "player_moved", movement, velocity, last_input)
 
+remotesync func _on_main_player_picked_up_item(item_path: String) -> void:
+	"""Called when the main player sends a request to pick up an item."""
+	rpc_id(1, "player_picked_up_item", item_path)
+
+remotesync func _on_main_player_dropped_item() -> void:
+	"""Called when the main player sends a request to drop an item."""
+	rpc_id(1, "player_dropped_item")
```
vainiovano: these functions shouldn't be `remotesync` if they are only called locally

src/assets/main/main.gd
```diff
@@ -157,6 +160,46 @@ func _on_main_player_moved(movement: Vector2, velocity: Vector2, last_input: int
...
+
+remotesync func player_picked_up_item(item_path: String) -> void:
+	"""Called by the server; RPCs all clients to have a player pick up an item."""
```
vainiovano: "Called by the server" means that the function is rpc'd by the server and executed by the clients in https://github.com/opensuspect/opensuspect/blob/2ff9d8997099a885bceb9445f1e5963c69d1e439/src/assets/main/main.gd#L210-L213

I think that saying "called by client/server" is a bit unclear because it can mean multiple things. At least its usage in the code should be consistent.

vainiovano (dismissed):

src/project.godot
```diff
@@ -9,6 +9,20 @@
...
+"language": "GDScript",
+"path": "res://assets/player/item_handler.gd"
+}, {
+"base": "KinematicBody2D",
+"class": "Player",
+"language": "GDScript",
+"path": "res://assets/player/player.gd"
 "base": "Control",
 "class": "ControlBase",
 "language": "GDScript",
```
TheSecondReal0: Add a line with ```}, {``` right after line 25, the class isn't being closed correctly so the Godot project can't be opened.

TheRyeGuyWhoWillNowDie (changes requested):
tell us when it's not broken

TheRyeGuyWhoWillNowDie (dismissed):
fixed now cool

src/assets/main/main.gd
jngo102: What do you recommend we call them?

src/assets/main/main.gd
jngo102: Changed.

src/assets/main/main.gd
vainiovano: "Runs on the server" sounds good

TheRyeGuyWhoWillNowDie (dismissed):
very yes

src/assets/main/main.gd
jngo102: Renamed comments.

TheSecondReal0 (dismissed):
Trying to host or join a game on an exported client causes the game to crash. The appearance editor seems bugged as well, it looks like this:

image

The animations still work, but the textures remain white. Trying to change any setting crashes the game (including skin color). For some reason the crash and editor bug only happen when the game isn't running in debug mode.

Other than this, could you make it so that you don't have to leave and reenter an item's radius to pick it back up?

TheSecondReal0 (changes requested):
Trying to host or join a game on an exported client causes the game to crash. The appearance editor seems bugged as well, it looks like this:

appearance-bug

The animations still work, but the textures remain white. Trying to change any setting crashes the game (including skin color). For some reason these issues only occur when the game isn't running in debug mode.

Other than this, could you make it so that you don't have to leave and reenter an item's radius to pick it back up?

nicemicro (dismissed):
I have two issues with this implementation at the moment:
1. Items are visible outside of vision range if not carried, and they appear when dropped.
2. Ghosts can pick up items and carry them. Items picked up by ghosts disappear, and reappear when dropped.

nicemicro:
> The animations still work, but the textures remain white. Trying to change any setting crashes the game (including skin color). For some reason these issues only occur when the game isn't running in debug mode.

By a cursory look it seems to me that the exported stuff doesn't contain the required png files, and also, there used to be a pck file alongside the binary, which doesn't exist in the zip file provided by the CI.
Moreover, the same is true for #261 for example.

src/assets/main/maps.gd
```diff
@@ -1,42 +1,47 @@
...
 		GameManager.State.Normal:
-			switchMap('test')
+			switch_map("test")
 
-func switchMap(newMap: String) -> void:
-	print('switchMap called for ', newMap)
-	if not maps.keys().has(newMap):
+func switch_map(new_map_name: String) -> void:
+	print("switch_map called for ", new_map_name)
+	var map_path: String = Helpers.find_file(new_map_name + ".tscn", maps_path)
```
TheSecondReal0, line 25: Replacing this line with:
```gdscript
var map_path: String = maps_path + new_map_name + "/" + new_map_name + ".tscn"
```
fixes the crash and allows the exported/non-debug version to run correctly. This only works if the folder and the scene have the same name, luckily for us that's already how it is.

TheSecondReal0 (changes requested):
While the problems with the appearance editor are not caused by this PR, there is a crash that occurs when you try to host or join a game on an exported/non-debug version. I've narrowed it down to being an issue with how maps.gd is finding the map file path. ```Helpers.find_file()``` seems to cause the crash.

Another minor thing is that you can't drop an item and immediately pick it back up, you have to walk far enough away so that the item can re-enter the item radius.

TheSecondReal0:
from @nicemicro
> By a cursory look it seems to me that the exported stuff doesn't contain the required png files, and also, there used to be a pck file alongside the binary, which doesn't exist in the zip file provided by the CI.
> Moreover, the same is true for #261 for example.

The PCK isn't created because it's set to include it inside the executable

nicemicro:
This work has been continued in a different PR due to the merge conflicts that arose here.

PR #255, opened 2020-12-08
State changed priority signals

TheSecondReal0:
* Added a new signal to the game manager, ```state_changed_priority```. It is identical to ```state_changed```, except it also passes a priority int.
## HOW IT WORKS
When the game state is changed, GameManager will emit ```state_changed_priority``` several times, with incrementing priority. The first time it is emitted the priority will be 0, the next it will be 1, etc.. No additional network syncing is added.
## HOW TO USE
To use this new signal you connect it to a function in your script. Example:
```gdscript
func _ready():
	set_network_master(1)
	GameManager.connect("state_changed_priority", self, "state_changed_priority")

func state_changed_priority(old_state, new_state, priority):
	# we want to do something on priority 2
	if priority != 2:
		return
	match new_state:
		# we want to do something when the game transitions to the Normal state
		GameManager.State.Normal:
			do_thing()
```
## NOTES
The normal ```state_changed``` signal is emitted after all of the priority signals.

Each priority level is used for a certain thing: 
### Priority 0
Switching main scenes in ```scenemanager.gd``` and toggling accept/refuse connections in ```network.gd```

Priority 0 should be reserved for cleanup and extremely essential functions.
### Priority 1
Switching map in ```maps.gd```
### Priority 2
Assigning roles in ```playermanager.gd```
### Priority 3-4
Currently unused.
### Priority 5
Spawning players in ```main.gd```

### DIFF
--- a/src/assets/autoload/gamemanager.gd
+++ b/src/assets/autoload/gamemanager.gd
@@ -9,9 +9,12 @@ const TRANSITIONS = {
 	State.Normal: [State.Lobby, State.Start],
 }
 
+var priority_amount: int = 6
+
 #signals that help sync the gamestate
 #can be connected to from anywhere with GameManager.connect("<signal name>", self, "<function name>")
-signal state_changed
+signal state_changed(old_state, new_state)
+signal state_changed_priority(old_state, new_state, priority)
 
 func _ready():
 	set_network_master(1)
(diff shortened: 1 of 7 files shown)


nicemicro (dismissed):
Code looks clean, and it is ready to merge, doesn't seem to break any unbroken functionality.
However, please change the version variable, as this PR introduces very significant change to the plumbing.

nicemicro (approved):
Code looks clean, doesn't break existing functionality when testing on one computer.

PR #257, opened 2020-12-08
Task system now functional/usable

### LINKED ISSUE #103: Task management system

TheSecondReal0:
* Tasks created with the InteractTask resource now work
* Tasks can trigger map interactions upon completion
* Task UIs can send data through the task system to be used in output interactions

## HOW TO USE
1. Configure the InteractTask resource. Refer to the [interaction resources wiki page](https://github.com/opensuspect/opensuspect/wiki/Interaction-Resources)
2. Create a UI that has a script that extends one of the task UI classes (```ControlTask```, ```PopupTask```, ```PopupPanelTask```, or ```WindowDialogTask```)
3. Add whatever logic you want to your task
4. When you want the task to be completed, call ```.complete_task()``` (the dot MUST be there). You can pass a dictionary in ```.complete_task()``` and its data will used when the output interactions are being executed. 

## HOW IT WORKS
The task manager is a central source/organizer for the task resources, which largely keep track of themselves. When the InteractTask resource is initialized, it will tell the task manager to register them. The task manager will give them their task ID and some data to keep track of (task state, players assigned, etc.). 

When someone interacts with the task resource, it will use its InteractUI resource to open a UI, and pass all of its data through it. The UI system should try to open a UI that inherits from a task UI class. The task UI class will double check that the task is valid, and then agree to opening the UI. 

When the task is completed (when ```.complete_task()``` is called) the task UI class will call ```complete_task()``` on the TaskManager, which will double check the validity of the request and then forward it to the task resource. The task resource will do whatever it needs to do, most likely triggering map interactions or nothing.

## TODO
* Task assignment
* Task resources still aren't able to discern which player is interacting with them
* Network syncing
* Condition dependent tasks (the state of something on the map, maybe a door being open). This is low priority because this is not very well defined. Somehow tasks will have to sense items, I don't know how this will happen yet.

## NOTES
* Clockset task in the top of the test map has been updated to use the task system
* The ```ui_data``` variable used in UI scripts is now built into the base classes, and by extension the task UI classes

### DIFF
--- a/src/addons/opensusinteraction/resources/interact/interact.tres
+++ b/src/addons/opensusinteraction/resources/interact/interact.tres
@@ -15,11 +15,11 @@ interact_data = {
 
 }
 
-[sub_resource type="Resource" id=2]
+[sub_resource type="Resource" id=4]
 resource_local_to_scene = true
 resource_name = "InteractUI"
 script = ExtResource( 4 )
-ui_name = "clockset"
+ui_name = ""
 ui_data = {
 
 }
@@ -31,7 +31,7 @@ resource_local_to_scene = true
 resource_name = "InteractTask"
 script = ExtResource( 5 )
 task_text = ""
-ui_resource = SubResource( 2 )
+ui_resource = SubResource( 4 )
 outputs/toggle_map_interactions = false
 
 [resource]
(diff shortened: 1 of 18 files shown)


TheRyeGuyWhoWillNowDie (approved):
no issues here

PR #264, opened 2020-12-15
Added network syncing to map interactions

### LINKED ISSUE #104: Map interaction system
A system that allows the map to be interacted with. This would manage stuff like opening doors, and other map interactions. Features to include:
- Store a list of dynamic map objects (such as doors)
- ~~Be able to receive signals from other systems, such as the task management system~~
- ~~Be able to change the state of various objects on the map~~

TheSecondReal0:
## GENERAL CHANGES
* You can now pass a bool in ```interact_with()``` to choose whether or not to sync the interaction over the network. It is the 4th input and optional.
* mapmanager.gd now uses ```push_error()``` instead of ```print()``` to display errors

## NEW HELPERS
```gdscript
var node = get_node("MyButton")

# returns path to node from root AKA absolute path
# ex. output: "Main/uicontroller/MyButton"
var absolute_path = Helpers.get_absolute_path_to(node)

# gets node from absolute path
# ex. output: [Button:1853]
Helpers.get_node_from_root(absolute_path)

# gets node from absolute path, if node does not exist returns null
Helpers.get_node_or_null_from_root(absolute_path)

# a property path points to a property (variable) within a node
# this path is pointing to the name variable inside the Button node
var property_path = Helpers.get_absolute_path_to(node) + ":name" # "Main/uicontroller/Button:name"

# returns value of the property that the property path points to
# ex. output: "MyButton"
Helpers.get_node_property_from_root(property_path)
```

## NOTES
* Nodes sent in ```interact_data``` will be retained when sent over network, but resources will not
* Clients cannot send map interactions to be synced. This is disabled for now because it is hilariously easy to exploit and I don't know how you would verify a map interaction

### DIFF
--- a/src/addons/opensusinteraction/resources/interactmap/interactmap.gd
+++ b/src/addons/opensusinteraction/resources/interactmap/interactmap.gd
@@ -11,6 +11,8 @@ var attached_to: Node
 #data to pass to the UI node
 export(Dictionary) var interact_data
 
+var network_sync: bool = false
+
 var reported_interact_data: Dictionary = {}
 
 #called to execute the interaction this resource is customized for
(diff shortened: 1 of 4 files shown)


nicemicro (dismissed):
Code looks clean, game doesn't break.

nicemicro (approved):
Code clean, didn't see any obvious bugs. Approved.

PR #265, opened 2020-12-17
Task network sync

Damjan94:
Tasks now sync to the server.

Once #264 is merged, I will correct a few things that I can't do cleanly without that feature(if you wanted to spawn items via a resource, when task is done, it would not work)


src/addons/opensusinteraction/resources/interacttask/interacttask.gd
```diff
@@ -29,12 +29,12 @@ var attached_to: Node
 
 #assigned at runtime when registered by TaskManager
 var task_id: int
-var task_data: Dictionary = {}
-
+var _task_data: Dictionary = {}
```
TheSecondReal0: What is the rationale for changing this variable's name?

TheSecondReal0 (changes requested):
I haven't been able to test it yet, but I've looked at the code. It looks like most task functions require a network/player id as an input, how does this work for global tasks? Maybe make them optional inputs?

src/addons/opensusinteraction/resources/interacttask/interacttask.gd
```diff
@@ -33,7 +35,9 @@ var _task_data: Dictionary = {}
 var task_data_player: Dictionary = {}
 var task_registered: bool = false
 
-func complete_task(player_id: int, data: Dictionary = {}) -> bool:
+func complete_task(	player_id: int = TaskManager.GLOBAL_TASK_ID,
```
TheSecondReal0: I personally would put ```player_id``` after ```data```, it doesn't really matter though. Maybe you could include the player data inside the task data dict (I think the keys are already generated).

src/assets/autoload/taskgenerators.gd
```diff
@@ -0,0 +1,30 @@
+extends Node
+
+#the task data(eg. currentTime, targetTime) should be set server side
+#these generators should generate new data every time they are called
```
TheSecondReal0: I'm a bit weary about adding a new autoload. It looks like this is mainly generating a starting position for the clockset task (and theoretically other tasks), I think it would be better to try to bake this into the interact task as part of the editor UI.

src/assets/autoload/taskmanager.gd
```diff
@@ -28,41 +29,88 @@ func _ready():
...
+puppet func receive_task_data(task_text: String, task_data: Dictionary):
+	var task_id = get_task_id_by_name(task_text)
+	if not does_task_exist(task_id):
+		assert(false)
+		return
+	emit_signal("receive_task_data", task_id, task_data)
 
+#removes parameters from task data that shouldn't be sent over network
+func networkfy_task_data(task_data: Dictionary) -> Dictionary:
+	var keys_to_erase = ["task_id", "task_outputs", "attached_node", "resource"]
```
TheSecondReal0: I'm not sure why the task ID is getting filtered out, shouldn't you be using it to refer to the task?

src/assets/common/classes/ui/task/controltask.gd
```diff
@@ -2,21 +2,53 @@ extends ControlBase
...
 
+func _ready():
+	#warning-ignore:return_value_discarded
+	TaskManager.connect("task_completed", self, "_on_task_completed")
+	#warning-ignore:return_value_discarded
+	TaskManager.connect("receive_task_data", self, "_on_received_task_data")
+
 func complete_task(data: Dictionary = {}):
-	if TaskManager.complete_task(ui_data["task_id"], data):
+	TaskManager.rpc_id(1, 	"complete_task_remote", ui_data["task_text"],
```
TheSecondReal0: I think we should be using the task ID to refer to tasks, as 2 different tasks could have the same text.

Could you also switch it up so that this calls a function inside the task manager which executes the RPC? Putting it here makes the codebase more confusing.

TheSecondReal0 (dismissed):
I'm not sure why this new code is referring to tasks by their text instead of their ID. I see that IDs are filtered out when sending over the network, but I don't understand why. It seems that you'd want the IDs to be able to reliably refer to tasks?

src/assets/common/classes/ui/task/controltask.gd
Damjan94: yes, I agree that it should be calling a function inside the TaskManager.

src/assets/autoload/taskmanager.gd
Damjan94: because the task ID needs to be unique, and is most likely different on every game instance.
syncing the task ID between the clients and the server would require more code, so I opted for this approach, since two tasks are rarely going to need to have the same description, and even if they do, one could just append a character, or something...

src/assets/autoload/taskgenerators.gd
Damjan94: I don't think bloating anything task related, would be a good idea. because every task would need to generate it's own data, there would be a lot of code(maybe?)
I'm not sure what you mean by baking it into an editor ui...

src/addons/opensusinteraction/resources/interacttask/interacttask.gd
Damjan94: I will look into this

nicemicro (approved):
Works as intended, no unexpected behavior was observed, the game doesn't work.
Seems to have addressed all issues @TheSecondReal0 had with the previous versions.

PR #269, opened 2020-12-31
Map info resources

TheSecondReal0:
* Added MapInfo resource class which stores data about maps. Currently they only store name, description, thumbnail, and scene path. They can be found in ```src/assets/maps/mapinfo```
* Created MapInfo resources for existing maps
* Switched maps.gd to use new resources instead of a dictionary (technically still uses a dictionary, but it's updated at runtime based on the MapInfo resources)
* Added helpers for finding certain types of files in a directory (.png, .tres, etc.)

I discovered that ```_ready()``` isn't called before the game manager emits more state_changed signals. This forces maps.gd to load a preset map in ```_ready()```, breaking consistency in how maps are switched. I tried to connect to the ```state_changed_priority``` signal in ```_init()``` of maps.gd, but it causes a crash due to some coupling between main.gd and maps.gd. Currently, maps.gd emits a signal when it has spawnpoints ready which calls ```_on_maps_spawn()``` in main.gd. The crash can be fixed by removing ```get_tree()``` from that function, but it could create conflicts. Ideally, main.gd should query maps.gd for spawnpoints when it needs them.

Another way to solve this problem is to make sure the main scene is completely loaded (```_ready()``` has been called on all nodes) before GameManager emits more signals. Some useful info: [change scenes manually](https://docs.godotengine.org/en/stable/tutorials/misc/change_scenes_manually.html), [SceneTree documentation](https://docs.godotengine.org/en/stable/getting_started/step_by_step/scene_tree.html), and [this reddit post](https://www.reddit.com/r/godot/comments/j33usl/how_to_make_a_script_wait_for_a_scene_to_finish/).

EDIT: Manual scene switching implemented in #289


nicemicro (dismissed):
The code looks clean, and it works as intended.

TheSecondReal0:
Should the map info resource script be stored in ```assets/common```? Thinking about it now and it doesn't make much sense to store the resource script in the same folder as the maps.

nicemicro (approved):
It works, no issue with code quality.

PR #272, opened 2021-01-03
Character customization screen refactoring for more custom color selection

nicemicro:
In this PR I have

1. added a new scene which uses PNG bitmaps for color selection,
2. modified the character sprites to be compatible with the shaders / color customization,
3. refactored the whole appearance editor screen to be able to include the new options,
4. added a randomize button for random character generation.

I will make this PR live as soon as I have finalized the color maps (current ones are provisional).


vincens2005 (dismissed):
Looks good, the only think I think we might want to change in the future is that the skin color picker is missing a lot of shades. Other than that, everything looks good.

TheSecondReal0 (dismissed):
On an exported client the appearance editor doesn't seem to find the correct files, and you end up switching between files in ```src``` instead of outfit textures. My best guess is that it's something in ```getFilePaths()``` on line 69 (nice) in appearancemanager.gd. That function uses .png instead of .png.import like ```partFiles()```

nicemicro:
> On an exported client the appearance editor doesn't seem to find the correct files, and you end up switching between files in `src` instead of outfit textures. My best guess is that it's something in `getFilePaths()` on line 69 (nice) in appearancemanager.gd. That function uses .png instead of .png.import like `partFiles()`

The only thing didn't work on an exported project if I export myself on my PC, was the color maps due to loading using Image.load() function that is documented to break on exported projects.
I corrected that error. I think the export issue might be a CI setup issue.

nicemicro:
> On an exported client the appearance editor doesn't seem to find the correct files, and you end up switching between files in `src` instead of outfit textures. My best guess is that it's something in `getFilePaths()` on line 69 (nice) in appearancemanager.gd. That function uses .png instead of .png.import like `partFiles()`

Couldn't fix the issue, but found something that's borked during export, and fixed that. Maybe that issue could get its own PR.

src/assets/autoload/appearancemanager.gd
```diff
@@ -50,7 +187,7 @@ func savePlayerAppearance():
 	"""
 	Saves the visual appearance of the player to the file.
 	"""
-	pass
+	SaveLoadHandler.save_data(customization_path, my_customization)
 
 master func queryCustomization(id: int) -> void:
```
Damjan94: this should be just a normal function(without master keyword)
line 199 (func sendCustomizationToServer()) should have puppet keyword instead
204 (func receiveCustomizationFromClient) should have master keyword
228 (func receiveBulkCustomization) should be puppet too.

I know that this isn't your code, but I think these changes should be made somewhere

Damjan94 (approved):
all seems to be working fine, I left comments on the old code, I leave it up to you if you think it should be changed in this PR.

src/assets/autoload/appearancemanager.gd
nicemicro: No, actually this is my code, so I'll take a look and change it.

src/assets/autoload/appearancemanager.gd
```diff
@@ -5,22 +5,159 @@ appearance of players so when the actual player nodes (or other nodes
...
 signal apply_appearance(id)
 
 func _ready():
 	"""
 	When this script is ready, it loads the saved customization of the user.
 	"""
+	var color_map: Image
+	var texture: StreamTexture
+	for color_map_name in custom_color_files.keys():
+		color_map = Image.new()
```
Damjan94: this assignment is redundant

src/assets/ui/submenus/appearance_editor/color_selector.gd
```diff
@@ -0,0 +1,88 @@
...
+	colormap.texture = img_texture
+	setupSize()
+
+func setColorTo(x_rel, y_rel):
+	coord_rel.x = x_rel
+	coord_rel.y = y_rel
+	var max_x = self.rect_size.x - 1
+	var max_y = self.rect_size.y - 1 - color_map_y
+	var coord: Vector2
+	coord.x = int(x_rel / 1.0 / AppearanceManager.COLOR_XY * max_x)
```
Damjan94: what's the purpose of the '1.0' here?(please explain like you're explaining to someone who has skipped all of the math classes xD )

src/assets/ui/submenus/appearance_editor/color_selector.gd
```diff
@@ -0,0 +1,88 @@
...
+		color_map_image.unlock()
+		color_preview.color = pixel_color
+		customization["r"] = pixel_color.r
+		customization["g"] = pixel_color.g
+		customization["b"] = pixel_color.b
+		emit_signal("color_changed", customization, my_name)
+		if coords.y < max_y - 35:
+			preview_border.rect_position.y = 10
+		else:
+			preview_border.rect_position.y = -40
```
Damjan94, line 82: I feel it's a bit trippy how the box jumps to the top. I think it would be fine if it were to be drawn on top of the elements below it. But that's just my opinion. Leave it as is if no one else agrees with me, or if it's too much work to get it any other way.

Damjan94 (changes requested):

src/assets/ui/submenus/appearance_editor/color_selector.gd
vincens2005: It shouldn't make a difference, / 1.0 should be removed

nicemicro: it is to implicitly cast x_rel to float, so the division doesn't get automatically rounded down to zero while keeping the order of the operations so it is easy to see what these lines do.

src/assets/ui/submenus/appearance_editor/color_selector.gd
nicemicro: In the current form, due to the container used, the things below appear "on top of" the things above.
As in, if the color preview of the skintone goes down, it will go behind the color map of the hair color, which is worse, than this trippyness.

src/assets/autoload/appearancemanager.gd
nicemicro: removed.

src/assets/ui/submenus/appearance_editor/color_selector.gd
nicemicro: actually managed to fix it with the float keyword

src/assets/ui/submenus/appearance_editor/color_selector.gd
Damjan94: ok, fair enough.

Damjan94 (approved):

Damjan94 (approved):
approved again, after joining opensuspect team

PR #274, opened 2021-01-05
Improving UI system/InteractUI resources

TheSecondReal0:
InteractUI resources can now be used to open, update, instance, and close UIs instead of only being able to open UIs. Also added individual functions in interactui.gd to open, update, instance, and close UIs so you don't need multiple resources interacting with the same UI. 

Updating a UI only passes it new data and calls ```base_update()``` and ```update()``` on the UI (if the UI script has those methods).

I plan to switch the UI system to use file paths to UI scenes instead of a dictionary in uimanager.gd. I'm also probably going to move away from using custom types (with the plugin) and just use custom classes because they are better in almost every way (and custom types are probably getting removed from Godot soon). It would also keep interaction resources from showing up in random dropdowns meant for a specific type of resource.

### DIFF
--- a/src/assets/autoload/helpers.gd
+++ b/src/assets/autoload/helpers.gd
@@ -34,6 +34,7 @@ func map(in_value: float, in_value_min: float, in_value_max: float, out_value_mi
 
 func get_absolute_path_to(node: Node, subname: String = ""):
 	var path: String = get_tree().get_root().get_path_to(node)
+	path = "/root/" + path
 	if subname != "":
 		path = path + ":" + subname
 	return NodePath(path)
--- a/src/assets/autoload/helpers.gd
+++ b/src/assets/autoload/helpers.gd
@@ -53,3 +53,36 @@ func get_node_property_from_root(path: NodePath):
 	var node = get_node_from_root(path)
 	var subnames = path.get_concatenated_subnames()
 	return node.get_indexed(subnames)
+
+func object_has_method_with_args(object: Object, method: String, args: Array) -> bool:
+	var method_args: Array = get_object_method_arg_names(object, method)
+	for arg in args:
+		if not method_args.has(arg):
+			return false
+	return true
+
+func object_has_method_with_arg(object: Object, method: String, arg: String) -> bool:
+	var method_args: Array = get_object_method_arg_names(object, method)
+	var arg_names
+	return method_args.has(arg)
+
+func get_object_method_arg_amount(object: Object, method: String) -> int:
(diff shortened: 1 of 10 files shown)


src/assets/autoload/helpers.gd
```diff
@@ -52,3 +53,36 @@ func get_node_property_from_root(path: NodePath):
...
+	var method_args: Array = get_object_method_arg_names(object, method)
+	for arg in args:
+		if not method_args.has(arg):
+			return false
+	return true
+
+func object_has_method_with_arg(object: Object, method: String, arg: String) -> bool:
+	var method_args: Array = get_object_method_arg_names(object, method)
+	var arg_names
+	return method_args.has(arg)
```
nicemicro, line 67: Is it necessary? Can't just the previous function be called with an array with one string with it? Does GDscript allow for using [varname] to call an array with only one element, which is var, in it?

nicemicro (commented):
Not strictly related to this commit, but I observed that when someone is killed with a task window open, the task window doesn't close. If the game ends due to someone being killed, the task window keeps open even though we get switched back to the lobby.
Should these UI resources handle that behavior, or would it be a job of something else?

src/assets/autoload/helpers.gd
Damjan94: I think he wanted to have a clean call syntax. so to avoid writing
`object_has_method_with_args(myObj, "awesomeMethod", ["thePerfectArg"])`
this looks kind of nicer
`object_has_method_with_arg(myObj, "awesomeMethod", "thePerfectArg")`

but where do these methods get called from?
in any case, if this is the only issue, I think we should merge it, so that it doesn't become stale..

TheSecondReal0:
For the sake of decoupling, this should probably be handled somewhere else (probably in the script that handles dying). There should be a way to close all open menus, so I'll add that tomorrow/later today. Hopefully it would make it easier to close menus when you die, but I'm not sure how you would handle overlays that need to close? Maybe a signal from PlayerManager would work?

src/assets/autoload/helpers.gd
TheSecondReal0: You are correct that I wanted a clean call syntax 😄 

These functions aren't actually used anywhere, I was thinking about making it optional to include a `ui_data` input when you define `update()` in a UI script. They are completely unnecessary, and can be removed.

Damjan94 (approved):
Tested, works great

nicemicro (approved):

PR #275, opened 2021-01-06
Load save file from preset

vincens2005:
This is a workaround because I tried making an actual random appearance and it didn't work. I just made a preset that it loads instead.


berkersal (approved):
Works as expected

PR #279, opened 2021-01-10
Items and inventory based on #254

### LINKED ISSUE #194: Items and inventory

nicemicro:
It has been a long time since #254 has been PR'd, so I tried to merge the main branch into it, so we can proceed.


zapzoop0099:
The game is acutally working fine and debug messages are also nice. But i have found a loophole in the system pretty easy to be fixed but important.Actually using this loophole player can pick up or to be more precise drag 2 items in one place.
How to reproduce:-

- Hold any item 
- Go near to another item
- And start spamming the "E" key and run in any direction 

Using this a player can easily drag two items at once place.This was also present in the Base PR. It can be easily fixed by changing the rate at which the player can pick up item

nicemicro:
Multiple things are addressed, including @zapzoop0099 's concerns for the abusability of repeated dast pick up and drop of items.
The item highlighting has been revamped, now the item you can pick up with the "E" key is highlighted in green, others that you can select by the mouse are highlighted in yellow.

Damjan94 (changes requested):
apart from the merge conflict,
the infiltrator can kill, pick up an item, reload(while holding the item), and kill again(while holding the item)

nicemicro:
> apart from the merge conflict,
> the infiltrator can kill, pick up an item, reload(while holding the item), and kill again(while holding the item)

Thanks for the in depth testing, guess this will require a little more in-depth rewrite of some parts, I'll get be back to it as soon as possible.

src/assets/autoload/helpers.gd
```diff
@@ -8,13 +8,31 @@ func pick_random(array: Array):
...
 	var combined_string: String = ""
 	for index in range(len(string_array) - 1):
 		combined_string += string_array[index] + separator
 	combined_string += string_array[-1]
 	return combined_string
 
+func find_file(file_name: String, start_path: String = "res://", recursive: bool = true) -> String:
+	"""Find file 'file_name' starting at directory 'start_path' recursively."""
+	var directory := Directory.new()
+	assert(directory.open(start_path) == OK)
```
Damjan94, line 21: I still think that this should be replaced with
```
if directory.open(start_path) != OK:
    assert(false)
    return
```
as the asserts get removed in the release build

src/assets/autoload/helpers.gd
```diff
@@ -8,13 +8,31 @@ func pick_random(array: Array):
...
+			var recursive_file: String = find_file(file_name, start_path + ("" if start_path.ends_with("/") else "/") + file, recursive)
+			if recursive_file != "":
+				return recursive_file
+		file = directory.get_next()
+	return ""
+
 func list_directory(path: String, recursive: bool = false) -> Array:
+	"""Return an Array of all the files in directory 'path' recursively."""
 	var directory := Directory.new()
 	assert(directory.open(path) == OK)
```
Damjan94, line 37: same here, asserts won't work in  release builds

src/assets/main/main.gd
```diff
@@ -4,8 +4,8 @@ export (int) var MAX_PLAYERS = 10
 #!!!THIS IS IMPORTANT!!!
 #CHANGE THIS VARIABLE BY ONE EVERY COMMIT TO PREVENT OLD CLIENTS FROM TRYING TO CONNECT TO SERVERS!!!
 #A way to make up version number: year month date hour of editing this script
-var version = 210108
-var intruders = 0
+var version = 21012000
+var intruders = 0 # NiceMicro's question: is this value ever referenced anywhere?
```
Damjan94, line 8: no, according to `grep -r "intruders" ./`

src/assets/main/players.gd
```diff
@@ -143,5 +146,55 @@ puppetsync func player_killed(killer_id: int, killed_player_id: int) -> void:
...
+
+remotesync func player_picked_up_item(item_path: String) -> void:
+	"""Runs on the server; RPCs all clients to have a player pick up an item."""
+	if not get_tree().is_network_server():
+		return
+	var id: int = get_tree().get_rpc_sender_id()
+	if not players.keys().has(id):
+		return
+	# print("(players.gd/player_picked_up_item) player=", id, " item=", item_path)
+	rpc("pick_up_item", id, item_path)
```
Damjan94, line 176: Could you write a TODO, to check if the player trying to pick up an item is actually close enough to the item?

src/assets/player/item_handler.gd
```diff
@@ -0,0 +1,137 @@
...
+# The item that will be picked up if the player decides to do so
+var _target_item: Item
+# The current item in the player's hand
+var picked_up_item: Item
+
+func _ready() -> void:
+	player.get_node("DeathHandler").connect("dead", self, "_on_Player_dead")
+	pickup_timer.connect("timeout", self, "_on_timer_timeout")
+
+func _process(delta: float) -> void:
```
Damjan94, line 28: should this then be removed?

src/assets/autoload/helpers.gd
nicemicro: The use of any code that just searches for files in directories will be removed in the future.

src/assets/autoload/helpers.gd
nicemicro: Same answer.

src/assets/main/main.gd
nicemicro: Probably we'll have to revamp the roles system to make it a little more flexible with different teams and roles, so I'd leave dealing with this for that time.

src/assets/main/players.gd
nicemicro: Let's make an issue instead to include more server-side verification of actions instead of just adding a few TODO-s here and there in the code..

src/assets/player/item_handler.gd
nicemicro: I left it in for the time being, because we might use it for further improvements of the system.

src/assets/player/infiltrator.gd
```diff
@@ -18,12 +18,13 @@ signal stopped_reloading
...
-# Whether the infiltrator may kill or not
-var _killing_enabled: bool = true setget enable_killing, is_killing_enabled
-# If the infiltrator has executed a kill and has not reloaded yet; separate from
-# _killing_enabled as an infiltrator may be reloaded but an event may prevent
-# them from killing
-var _reloaded: bool = true setget set_reloaded, is_reloaded
+export (bool) var reload_only_empty := true
+
+# How many shots the infiltrator can take before needing to reload, and the shots left
+const max_shots: int = 1
```
Damjan94, line 24: The `infiltrator_hud.gd` catches the first `"kill"` signal, and disables the kill button, and enables the reload button.
when this value is set to more than 1, this makes it so that after the first reload the infiltrator is able to fire for one more time. After that the `infiltrator_hud.gd` switches to the reload button, but infiltrator is unable to reload(`reload_only_empty == true`)

The simplest fix I can think of is to decrement the `_shots_left` before emitting the `"kill"` signal, and passing that value in the signal. Then in `infiltrator_hud.gd` just return if the value is >0...

Damjan94 (changes requested):

Damjan94 (approved):

PR #280, opened 2021-01-11
Fix spawnpad's pixel blur

broccoder:
Since this is clearly a pixel art resource, it should be imported *without* anti-alias/filter.


broccoder:
Comparison of the resource, with and without filter option in import

with filter
image

without filter
image

how it looks in-game
image

nicemicro (approved):
Thanks,
even though it's just a placeholder asset, it's better if it looks good than if it looks bad.
@infinitygamer404 please feel free to join the development!

PR #284, opened 2021-01-14
Tasks presented in the taskinfo are less confusing

Damjan94:


nicemicro (approved):
Does what it says it does, improves the code quality.
Tested with the new "main" merged into it, no conflicts, good to go.

PR #286, opened 2021-01-15
Fixes the rolescreen crash(#271)

### LINKED ISSUE #271: Role screen crashes game on second created game
Way to reproduce:
1. Create a new game
2. Start the game
3. Go back to lobby
4. Go back to main menu
5. Create a new game
6. Start the game

Game crashes with error: Attempt to call function 'set_movement_disabled' in base 'null instance' on a null instance.
Where it occures: 
```
func _clean_up():
	for id in PlayerManager.players:
		PlayerManager.players[id].set_movement_disabled(false)
```
the PlayerManager.players is a dictionary with one element that is NULL

Damjan94:
Makes the rolescreen go thrugh the ui manager


nicemicro (changes requested):
TLDR: the instance that was server first, when tries to join a game hosted by an other instance, breaks on rolescreen.

I can still get a crash on the rolescreen:
- run OpenSuspect in 4 instances
- create a game on instance 1
- join all other clients, start, back to lobby, leave the game.
- create a game on instance 1
- join all other clients, start -> rolescreen works, back to lobby, leave the game.
- create a game on instance 2
- join all other clients, start -> rolescreen breaks on instance 1.

The crash happens in rolescreen.gd:101 at function _create_info, with an` invalid get index on base 'Dictionary'`. For some reason, the player_sprite_collection didn't get populated properly.

TheSecondReal0:
This sounds like it has something to do with network info (peers + names) not being cleared when you return to the main menu. Maybe adding a cleanup phase to network.gd would help?

IIRC roles aren't cleared either, but that shouldn't be causing any problems because they are reset each time roles are assigned/received.

src/assets/ui/hud/defaulthud/rolescreen/rolescreen.gd
```diff
@@ -11,27 +11,28 @@ func _ready():
...
 		return
+	reset()
+	populate_player_info_container(player_roles)
 
+func populate_player_info_container(player_roles: Dictionary):
 	display_timer.start()
-	PlayerManager.inMenu = true
-	""" 
+
+  """
```
nicemicro: That three quotation marks with the spaces in front of them instead of a tab makes Godot rage quit if running the game from the editor.
Moreover, I don't think it's good form to have the TODO in a string like this, let's just make it a normal comment with #

nicemicro (changes requested):
If I correct the abovementioned spaces to a tab character, the game works.

src/assets/ui/hud/defaulthud/rolescreen/rolescreen.gd
Damjan94: I resolved the merge conflict here, on github so I didn't notice the spacing...
good thing you caught that :)

nicemicro (dismissed):
Good, works perfectly, code is clean.

nicemicro (approved):

PR #289, opened 2021-01-22
Scene switching/transition fix

TheSecondReal0:
Without manual scene loading, a transition that includes switching scenes will be unreliable. Currently, when you transition to lobby from start, Godot emits signals in reverse order and then switches scenes. So priority 5 is emitted before 4, 4 before 3, etc. THEN loads the scene, which means anything related to nodes in the new scene doesn't happen (notably player/map spawning). This has caused some weirdness and we've implemented a few workarounds because of it (automatically spawning a player and loading map in `_ready()` functions instead of waiting for the corresponding signal). 

Hopefully this PR will make transitions much more consistent.


nicemicro (approved):
Code is clean, works like a charm.

PR #293, opened 2021-01-30
Container tasks

### LINKED ISSUE #196: Tasks system
Implement a flexible tasks system. This is a large system to implement, might take multiple people to tackle it, or one very dedicated soul.

Should have:
- Task assignment, based on role / randomness
- Interaction between player, task and map
- For exact details, see design document, linked below

An example task can be seen at this PR: https://github.com/opensuspect/opensuspect/pull/192
See https://github.com/opensuspect/design-document/blob/main/task_base.md.

zapzoop0099:
A container task as intended by the design document 
Currently it is fully functional/completed **From Game Perspective** but enough for getting items from the box

This PR provides:-

- A full sync between the task
- A location from where player can pick up the items


src/assets/autoload/uimanager.gd
```diff
@@ -85,6 +87,11 @@ func free_ui(ui_name: String):
 		push_error("free_ui() called with invalid ui name " + ui_name)
 	emit_signal("free_ui", ui_name)
 
+func pre_ins(ui_name:String):
```
Damjan94: what does this function do(maybe a more descriptive name)?

src/assets/main/maps.gd
```diff
@@ -68,6 +68,8 @@ func switchMap(newMap: String) -> void:
 	# actual current map to position 0 manually.
 	move_child(mapClone, 0)
 	emit_signal("spawn", getSpawnPoints())
+	if newMap == "Test":
+		init_all_tasks()
```
Damjan94: This could be moved below line 128 in uimnager, because it is ui based, and that's where I have put `open_ui("rolescreen")`.

wait for other peoples input, before you change this one, as I'm not 100% sure myself

src/assets/maps/test/Container.gd
```diff
@@ -0,0 +1,45 @@
...
+	pass 
+
+func reset() -> void: #Resets the node
+	interacting = false
+	pressed = false
+	interactor.clear()
+	
+func register(body) -> void:#Register a body
+	interactor[body.id] = body.get_path()
+	interactor["bool"] = true
```
Damjan94: What does this bool do? more descriptive names can never hurt

src/assets/maps/test/Container.gd
```diff
@@ -0,0 +1,45 @@
...
+		UIManager.open_ui("container", interactor)
+		#interact(interactor)
+
+func _on_Container_body_entered(_body):#Check wether the body is player or not and record the player
+	for bodies in get_overlapping_bodies():
+		if bodies.is_in_group("players"):
+			if interactor.empty():
+				register(bodies)
+			else:
+				return
```
Damjan94: this can be better structured as
```
	# Can't register a new body, if there is an already registered body
	if is_body_registered():
		return

	for bodies in get_overlapping_bodies():
		if not bodies.is_in_group("players"):
			continue

		register(bodies)
		return

func is_body_registered() -> bool:
	return interactor.size() > 0
```

src/assets/ui/tasks/container/container.gd
```diff
@@ -0,0 +1,83 @@
...
+
+puppetsync func pass_data(data:Dictionary):#Pass data to child
+	for slot in grid.get_children():
+		slot.ui_data = data
+		
+func erase_data():#Erase data from child
+	for slot in grid.get_children():
+		slot.ui_data.clear()
+
+puppetsync func erase_child():#erases all child
```
Damjan94: `erase_children`, plural of child is children

src/assets/ui/tasks/container/container.gd
```diff
@@ -0,0 +1,83 @@
...
+remotesync func pass_data_server(data:Dictionary):
+	if not get_tree().is_network_server():
+		return
+	rpc("pass_data", data)
+	
+remotesync func reset_server():
+	if not get_tree().is_network_server():
+		return
+	rpc("reset")
+remotesync func erase_child_server():
```
Damjan94: don't forget to change it to `children` here, too... English is sometimes complicated like that(or is it just me? :D)...

src/assets/ui/uicontroller/uicontroller.gd
```diff
@@ -127,3 +130,16 @@ func get_child_node_names() -> Array:
 	for i in get_children():
 		name_list.append(i.name)
 	return name_list
+
+func pre_ins(ui_name: String):
```
Damjan94: why not do
```
func pre_ins(ui_name: String):
	instance_ui(ui_name)
	close_ui(ui_name)
```

src/assets/ui/tasks/container/container.gd
```diff
@@ -0,0 +1,83 @@
...
+
+func _ready():
+#When the first player interacts then called this func to all others so everyone get the same scene set
+	if grid.get_child(0) == null:
+		rpc_id(1,"set_slot_server")
+#Connects with GameManager to record game transition
+	GameManager.connect('state_changed', self, '_on_state_changed')
+
+
+func _process(_delta):#Called every frame to check interaction status
```
Damjan94: is there no way to change this to be event based?

Damjan94 (changes requested):
This is the first change request. I just glanced over the code, and made suggestions.
I also checked out the code, and it seems that if I take all of the items out of the container in one go(ie, I only open the window once), the items end up being on the ground, and when I try to pick them up, they teleport to the centare(0,0) of the map...
I will make another review if I find more bugs.

nicemicro (changes requested):
I have not checked the codebase yet, I trust @Damjan94 that he did a thorough job and I'll look at it after his recommendations have been implemented.
However to expand on the bug @Damjan94 describes in their comment, it happens wit every item you pick up from the storage. Steps to reproduce:
1. take an item out of the storage.
2. walk away and drop the item on the ground.
3. try to pick up the item again
4. watch the item move to the center of the map
5. the item can not be picked up again.

src/assets/items/item.gd
```diff
@@ -34,7 +35,11 @@ func picked_up() -> void:
 	"""Item is picked up."""
 	set_collision_layer_bit(4, false)
 	being_held = true
-	map_items.remove_child(self)
+	if item_from_container:
+		var path = get_tree().get_root().get_node(item_location)
+		path.remove_child(self)
```
Damjan94, line 39: the bug @nicemicro and I found(where items get teleported to (0,0) coordinate) is because item_from_container is never set to false. you could set it to false here...

Damjan94 (changes requested):
if you add `assert(picked_up_item == null)` before [this line](https://github.com/zapzoop0099/opensuspect/blob/8d62ec3c1f97e7b5d43569f25027e9c3cf4b2735/src/assets/player/item_handler.gd#L79), sometimes, when there's a lot of items around the container with the batteries, you end up with a perma item in your hand, that you can't drop.
image
actually, there don't need to be any items around the container. It sometimes will just put two items if it doesn't close the ui after you take away the first item, and you click to take away another item

src/assets/player/player.gd
```diff
@@ -77,8 +78,11 @@ func _ready():
 	customizePlayer(id)
 
 func _input(event: InputEvent) -> void:
-	# Item Handler does not receive input as a child of a Viewport
-	item_handler._input(event)
+	if can_pickup == true:
+		# Item Handler does not receive input as a child of a Viewport
+		item_handler._input(event)
```
nicemicro, line 83: the item_handler should do all the different types of checking whether it is possible to pick up an item or not.
Having a bool for this purpose is not good, because many different events can block pick up, and if two blocking events occur simultaneously, and one is over, that would maybe switch the bool back to true, even though the other blocking event still did not pass.
So instead of having a boolean, just add the additional conditions to `item_handler._input()`.

src/assets/ui/tasks/container/container.gd
```diff
@@ -0,0 +1,83 @@
...
+#When the first player interacts then called this func to all others so everyone get the same scene set
+	if grid.get_child(0) == null:
+		rpc_id(1,"set_slot_server")
+#Connects with GameManager to record game transition
+# warning-ignore:return_value_discarded
+	GameManager.connect('state_changed', self, '_on_state_changed')
+
+func _process(_delta):#Called every frame to check interaction status
+	if not ui_data.empty():
+		rpc_id(1, "pass_data_server", ui_data)
```
nicemicro: I also don't understand, what happens if the ui_data becomes empty, when it was not empty before? does it not RPC the server in that case?
Moreover, when this RPC is called, the server will send the RPC to all clients, so it ends up setting it to itself again. Is this intended?

src/assets/ui/tasks/container/container.gd
```diff
@@ -0,0 +1,83 @@
...
+	for slot in grid.get_children():
+		slot.ui_data.clear()
+
+puppetsync func erase_children() -> void:#erases all child
+	for child in grid.get_children():
+		child.queue_free()
+
+puppetsync func reset() -> void:#Clears all the data
+	erase_data()
+	ui_data.clear()
```
nicemicro: shouldn't we check here whether the RPC is coming for the server?

nicemicro (changes requested):
I feel like the container.gd is doing too much with respect to synchronizing the data.
I would be more happy if we used an autoload like taskmanager.gd or rather a script attached to main.tscn to handle these kind of synchronization behaviors, as I would think in the future when we have multiple different tasks, having a common back end would be beneficial.

PR #296, opened 2021-02-11
Keep camera centered on player

monban:
Looking for some feedback on this idea, I think it's nicer to have the camera centered on the player instead of waiting for them to walk to the edge of the screen. Gives you a better sense of what's around you, more "awareness", feels less claustrophobic.


nicemicro (approved):
I like it this way better.

PR #298, opened 2021-02-11
Pick-up highlighting fixed

### LINKED ISSUE #294: Items green right after swapping them
get together 3 items like so
 
image

As soon as you swap one, the other one becomes immediately green.
Expected behavior would be that no items are highlighted green, if the player is unable to pick them.

nicemicro:
the pick up highlighting is fixed to be aligned with actual pick upability.

should close #294

### DIFF
--- a/src/assets/main/players.gd
+++ b/src/assets/main/players.gd
@@ -172,6 +172,11 @@ remotesync func player_picked_up_item(item_path: String) -> void:
 	var id: int = get_tree().get_rpc_sender_id()
 	if not players.keys().has(id):
 		return
+	# Before RPC'ing the other clients, we test whether the item can be picked up.
+	# Currently, it is only checking if pickup is enabled for the player, but
+	# doesn't check distance etc.
+	if not players[id].item_handler.isPickupEnabled():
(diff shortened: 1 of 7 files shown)


src/assets/main/players.gd
```diff
@@ -172,6 +172,11 @@ remotesync func player_picked_up_item(item_path: String) -> void:
 	var id: int = get_tree().get_rpc_sender_id()
 	if not players.keys().has(id):
 		return
+	# Before RPC'ing the other clients, we test whether the item can be picked up.
+	# Currently, it is only checking if pickup is enabled for the player, but
+	# doesn't check distance etc.
+	if not players[id].item_handler.isPickupEnabled():
```
Damjan94, line 178: for some reason, while holding an item, if you set a breakpoint on this line, the item lights up green as soon as it's dropped...

nicemicro: Can not reproduce if the program runs.
I think the breakpoint only creates the issue due to the timer expiring in the meantime.

Damjan94 (approved):

PR #300, opened 2021-02-13
Custom task logic via extending InteractTask

### LINKED ISSUE #196: Tasks system

### LINKED ISSUE #198: Maintenance tasks
A task type, that is never "started" or "finished", but constantly emits an output value based on some random input and some setting in the task interface (check https://github.com/opensuspect/design-document/blob/main/general_gameplay.md for reference).
The input value(s) should be constantly managed by the server, and task, where the setting is available, should provide the same interface with all players, so that should also be managed by the server. The output value should be also available to tasks that depend on the maintenance task, and certain values should trigger events such as warning sirens or even win / loss conditions.

There is a mock implementation of a gas valve task where the output should be regulated with a randomly changing input in [PR#192](https://github.com/opensuspect/opensuspect/pull/192).

TheSecondReal0:
* Added a bunch of virtual functions meant to be overridden to interacttask.gd
* Added functions in both interacttask.gd and the task manager to facilitate syncing task resource scripts (they are basically proxies for rpc and rset)
* Ported the clockset task connected to the standbutton to use a script extending InteractTask for logic (tasks are still completed via the UI class)
* Task networking supports RPC modes
* Task resources now tell the task manager when they are complete

I'll type up some documentation at some point, to see an example of using the InteractTask resource for logic look at the clockset task (resource script is at ```res://assets/ui/tasks/clockset/clocksetinteracttask.gd```)

TODO:
* ~~Have task resources tell the task manager when they are completed~~
* ~~Add some security to task resource rpc and rset, currently any peer can do whatever they want~~
* ~~Maybe change task manager to use task IDs instead of task_info dicts, @Damjan94 what is your opinion on this?~~ Necessary to make sure you know which player(s) the call is referring to

NOTES:
* You must make your InteractTask extending script a tool script, otherwise the custom editor properties won't work


src/addons/opensusinteraction/resources/interacttask/interacttask.gd
```diff
@@ -87,14 +108,54 @@ func assign_player(player_id: int = TaskManager.GLOBAL_TASK_PLAYER_ID):
...
+func task_rpc(function: String, args: Array):
+	TaskManager.task_rpc(function, args, task_id)
+
+func task_rpc_id(id: int, function: String, args: Array):
+	TaskManager.task_rpc_id(id, function, args, task_id)
+
+func receive_task_rpc(function: String, args: Array):
+	# using callv instead of call because it will translate the array into
+	# 	individual arguments
+	callv(function, args)
```
nicemicro: couldn't it be done so that task resources can only RPC the server, and can only accept RPCs from the server?

nicemicro, line 156: sholdn't it be enough for the task resource to only ever RPC the server, and then the server only validates the data and then receives the data only from the server?

nicemicro (commented):
Quite cool!

src/addons/opensusinteraction/resources/interacttask/interacttask.gd
TheSecondReal0: Yeah, it could be done that way. I set it up like this so each task can fully customize their networking stuff 😃

src/assets/ui/tasks/clockset/clockset.gd
```diff
@@ -14,38 +10,54 @@ func _ready():
...
+	get_res().set_current_time(time)
+
+func getTargetTime() -> int:
+	return get_res().get_target_time()
+
+func getCurrentTime() -> int:
+	return get_res().get_current_time()
+
+func get_res() -> Resource:
+	return TaskManager.get_task_resource(ui_data[TaskManager.TASK_ID_KEY])
```
nicemicro, line 60: I am trying to figure out where is the ui_data dictionary gets populated with the correct task ID, but I can't figure it out. Please enlighten me.

src/assets/ui/tasks/clockset/clocksetinteracttask.tres
```diff
@@ -0,0 +1,25 @@
...
+advanced/free_on_close = false
+
+[resource]
+resource_local_to_scene = true
+script = ExtResource( 1 )
+task_text = "clockset task"
+ui_resource = SubResource( 1 )
+outputs/toggle_map_interactions = true
+outputs/output_map_interactions = [  ]
+is_task_global = true
```
nicemicro, line 25: What is the purpose of this file?
I saw that the map features use the `interactpointclockset.tres` and `standbuttonclockeset.tres` resources as Interact Resource in their respective points, however those resources are linked to `clocksetinteracttask.gd` as their script. However, this tres file seems to be referenced nowhere.

nicemicro (changes requested):
Besides the questions with respect to the code, the following bugs need to be addressed:
1. The task list fails to color the completed tasks green.
2. When the clockset task (standbutton) is completed, the text showing the time next to the standbutton is only changed on the other players' map if they stand on the button.
3. On the interact button version, the text is only updated if the other player goes there and clicks on the task test button.

I'd suggest that 2. and 3. should be addressed that every time there is an update broadcasted to all clients, the script should update the text.

src/assets/ui/tasks/clockset/clockset.gd
TheSecondReal0: It is updated by uicontroller.gd right before it's opened

src/assets/ui/tasks/clockset/clocksetinteracttask.tres
TheSecondReal0: It was mostly for testing, and so I could make a copy of it to make the resource files used in the test map. Just putting that resource into the map either did really weird stuff (like adding a copy of the script inside test.tres) or I couldn't give them unique task text (despite them both being local to scene).

TheSecondReal0:
@nicemicro I think bugs 2 and 3 are caused by incorrect syncing of the clockset task, which would explain why it works perfectly if you complete the task on the host. I'll fix the syncing issues, and I'll make sure the task state is actually transitioned as @Damjan94 brought up in the Matrix chat

TheSecondReal0:
Bugs 2 and 3 were actually caused by ```attempt_complete_task()``` being called normally instead of via RPC

nicemicro (dismissed):

Damjan94 (dismissed):
If you make any task not global
image
and try to complete it, the game crashes...

src/addons/opensusinteraction/resources/interacttask/interacttask.gd
```diff
@@ -123,46 +216,185 @@ func gen_task_data() -> Dictionary:
...
+func transition(new_state: int, player_id: int) -> bool:
+	# to add custom behavior/checks before the state is officially changed
+	# if _transition() returns false, interpret it to mean the extending script
+	# 	doesn't want to transition
+	# to remove behavior after this point, you must override this function (make sure
+	# 	to emit the "transitioned" signal)
+	var virt_return = _transition(new_state, player_id)
+	if virt_return is bool and virt_return == false:
+		return false
+	var old_state = task_data_player[player_id]["state"]
```
nicemicro, line 270: the place where one of the weird bugs appear

nicemicro (dismissed):
I have found the following issues, please take a look.

1. Non-global tasks have the same final state (clock hands) and the states ("digital" clock) are synced through clients
2. Attempting to finish a non-global task on a client when the server is ran from the editor, results in an assertion error on the server in taskinfo.gd _on_task_completed() function, line 30
3. Attempting to finish a non-global task on a client when the client is ran from the editor doesn't cause any issues.
4. Attempting to finish a non-global task on a client not ran from the editor causes an error on a client that's ran from the editor in interacttask.gd transition() function, line 241 "Invalid get index (on base: Dictionary).
5. When all clients and the server are ran not from the editor, issues 2, 3 and 4 are not reproduced, however issue 1 causes the behavior that when someone finishes their clockset task, the second person who has the same personal task finishes the task immediately on opening it due to the state syncing.
6. If two players are attending to the same global task, and one changes the state, closes the window, the second player's display changes which is expected behavior. However, if one of the players is killed while their task window is open, they can still change the task state and when closing the window it gets synced across the network.

TheSecondReal0:
To respond to your review @nicemicro
1. This is because I wrote the clockset logic with only global tasks in mind, I'll alter it accordingly
2. That assertion should be removed, as it will trigger an error when the game is functioning correctly
3. This seems to just be a note
4. This might be caused by some weird syncing, I'll take a closer look in the future
5. This is also because the clockset logic is only written to be a global task, like 1
6. This should probably fall under another PR, since this deals with death behavior (there should probably be a signal from PlayerManager which the UI system would use to close UIs). There could also be a default check added in ```sync_task()``` for whether or not the main player is alive

nicemicro (changes requested):
I found the following issue:
1. The global clockset task have different different goal times on each clients.
2. When the first player opens the task window and closes it, the goal time gets set to the goal of the first player to everyone else.
3. The text next to the stand button shows the same time for everyone, but it only matches with the time on the client that was the last to join the party.
4. The text next to the stand button is only updated in case of successful completion of the task. Is it possible to update the related map item the same way when the task UI is updated, i.e. every time someone closes the task UI / changes some important parameter, so there could be some indication of the state of a task directly on the map in some cases?

TheSecondReal0:
* Addressed issues 1, 2, and I think 3 by adding player data syncing (sync occurs when the server sends task assignments)
* To address issue 4, I made it so that whenever times are synced on global clockset tasks it will update the map text, and personal tasks will update map text if it's assigned to you

nicemicro (approved):
This! Is! Approved!

PR #303, opened 2021-02-27
Quick fix to unassigned global tasks

TheSecondReal0:
This fixes a crash that occurs when you try to interact with a global task that was not actually assigned to the global task ID by the task manager. Not making it so global tasks are always assigned so that the game can pick and choose which global tasks to have in each round.


zapzoop0099 (approved):
It Fixes the crash regarding global task.

PR #306, opened 2021-03-16
Misc improvements to custom task logic

TheSecondReal0:
* Data passed when calling `interact()` on a task resource is now actually used
* No longer necessary to pass an empty array when calling `task_rpc()` or `task_rpc_id()` with no arguments


nicemicro (approved):
Doesn't seem to break anything, but adds important stuff to the task system back end.

PR #315, opened 2021-07-07
Replace multiline strings with comment blocks

vincens2005:
multiline strings are not removed by the JIT compiler, so they impact performance.


nicemicro (approved):
Good work.

PR #318, opened 2021-08-04
Fix ghost visiblility bug

### LINKED ISSUE #316: Ghosts are visible to everyone
An issue that has been introduced by probably one of the latest PRs is that the ghosts are visible to everyone.

vincens2005:
closes #316


nicemicro (approved):
Works as intended.

PR #322, opened 2021-11-20
Make player movement more smooth

vainiovano:
Current problems:
- The controlled player doesn't stop moving very quickly
- The server's position update sending rate is still too high


src/assets/player/player.gd
```diff
@@ -159,8 +161,9 @@ func run_physics(motion):
 		velocity.x = lerp(prev_velocity.x, 0, 0.17)
 	if velocity.y == 0:
 		velocity.y = lerp(prev_velocity.y, 0, 0.17)
-	# TODO: provide a delta value to this function and use it here
-	velocity = move_and_slide(velocity)
+	var collision = move_and_collide(velocity / 60)
```
nicemicro, line 164: what is the reason for this being divided by 60?

